PR #23, opened 2018-04-20
Fix cc empty initialiser and the multiple declaration in ufc.c

nate-sime:
Fix cc empty initialiser `= {}`
Remove the multiple declaration problem with more than one compiled ufl file `UFC_VERSION` in ufc.c.


garth-wells:
Did the docstring get added?

PR #26, opened 2018-05-03
Add #include<complex.h> to the include list

IgorBaratta:
Insert `#include<complex.h>` when the scalar type is defined as `double complex`.
This allows dolfinx to use the generated c code (without throwing an error), even though it it does not use complex scalars yet.

Alternatively, this could be done by the backend (more changes needed).

### DIFF
--- a/ffc/formatting.py
+++ b/ffc/formatting.py
@@ -184,10 +184,10 @@ def _generate_includes(includes, parameters):
     # s2 = external_includes - s
     includes_h = "\n".join(sorted(s_h)) + "\n" if s_h else ""
     includes_c = "\n".join(sorted(s_c)) + "\n" if s_c else ""
-    
+
     # This should really be set by the backend
     scalar_type = parameters.get("scalar_type")
     if scalar_type == "double complex":
-      includes_c += "#include <complex.h> \n"
-    
+        includes_c += "#include <complex.h> \n"
+
     return includes_h, includes_c
--- a/ffc/formatting.py
+++ b/ffc/formatting.py
@@ -82,8 +82,9 @@ def format_code(code, wrapper_code, prefix, parameters):
     code_c_pre += format_template["header_c"] % {"prefix": prefix}
 
     # Generate includes and add to preamble
-    includes_h = _generate_includes(includes, parameters)
+    includes_h, includes_c = _generate_includes(includes, parameters)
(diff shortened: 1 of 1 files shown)


ffc/formatting.py
```diff
@@ -177,6 +178,11 @@ def _generate_includes(includes, parameters):
...
     # s2 = external_includes - s
+    includes_h = "\n".join(sorted(s)) + "\n" if s else ""
 
-    includes = "\n".join(sorted(s)) + "\n" if s else ""
-    return includes
+    # This should really be set by the backend
+    if "scalar_type" in parameters.keys():
+        if parameters["scalar_type"] == "double complex":
+            includes_c = "#include<complex.h>" + "\n"
+
```
garth-wells: Maybe a bit more like:
```
scalar_type = parameters.get("scalar_type")
if scalar_type is "scalar_type": 
    includes_c = "#include<complex.h>\n"
```

ffc/formatting.py
```diff
@@ -177,6 +178,11 @@ def _generate_includes(includes, parameters):
     s = set(default_h_includes) | includes
 
     # s2 = external_includes - s
+    includes_h = "\n".join(sorted(s)) + "\n" if s else ""
```
garth-wells: Put on two lines.

ffc/formatting.py
chrisrichardson: you mean `if scalar_type is "double complex":`

garth-wells:
Is there a reason why `complex.h` isn't included in the header file? I would have thought the complex type would appear in the interface (function signatures).

ffc/formatting.py
IgorBaratta: Much better.
I'll keep `==` instead of `is`, otherwise the if would never be true.

IgorBaratta:
I was including `complex.h`  in header file, however the error persisted.
Maybe it's because the the complex type signature is not defined in the header file.
I'll look into it.

IgorBaratta:
I was including `complex.h` in the header file, however the error persisted.
Maybe it's because the the complex type signature is not defined in the header file.
I'll look into it.

IgorBaratta:
I'll work further on the complex case, including the signature of tabulate_tensor. And then open a new PR.

PR #31, opened 2018-06-01
Create interface for complex scalars

IgorBaratta:
The interface for the tabulation of the local element tensor contributions assumes that both the local tensor and coefficients are real-valued. 

The proposal is to create a more flexible interface by using `ufc_scalar` for both `A` and `w`, and then in _dolfin_ we could define for which datatype it refers. Where appropriate in dolfin use either of:
     `using ufc_scalar = double `
     `using ufc_scalar = std::complex<double>`

This changes allow dolfin to be compiled with petsc-complex and only small changes are required for the real scalars. Must be considered together with PR [#96](https://github.com/FEniCS/dolfinx/pull/96).

Obs.: In use in branch [complex-petsc](https://github.com/FEniCS/dolfinx/tree/igor/complex-petsc).


ffc/formatting.py
```diff
@@ -187,4 +192,22 @@ def _generate_includes(includes, parameters):
...
+    # Define the ufc_scalar type before including  the ufc header
+    # By default use double scalars
+    scalar_type = parameters.get("scalar_type")
+    if scalar_type == "double complex":
+        scalar = "#if defined(__cplusplus) \n" + \
+                 "#include <complex> \n" + \
+                 "typedef std::complex<double> ufc_scalar_t; \n" + \
+                 "#else \n" + "#include <complex.h> \n" + \
+                 "typedef double _Complex ufc_scalar_t; \n" + \
+                 "#endif \n"
```
blechta: AKA
```
scalar = "\n".join([
    "if defined(__cplusplus)",
    "#include <complex>",
    "typedef std::complex<double> ufc_scalar_t;",
    "#else",
    "#include <complex.h>",
    "typedef double _Complex ufc_scalar_t;",
    "#endif",
    "",
])
```

or

```
scalar = """if defined(__cplusplus)
#include <complex>
typedef std::complex<double> ufc_scalar_t;
#else
#include <complex.h>
typedef double _Complex ufc_scalar_t;
#endif
"""
```

ffc/formatting.py
```diff
@@ -80,6 +80,11 @@ def format_code(code, wrapper_code, prefix, parameters):
     code_h_pre += FORMAT_TEMPLATE["header_h"]
     code_c_pre += FORMAT_TEMPLATE["header_c"]
 
+    # Define ufc_scalar before including ufc.h
+    scalar_type = _define_scalar(parameters)
+    code_h_pre += scalar_type
+    code_c_pre += scalar_type
```
blechta: Why does the c file have the definition again? It does not include the header file?

ffc/formatting.py
IgorBaratta: Updated to option 1.

ffc/formatting.py
IgorBaratta: The header (h file) is not included in the c file. 
I don't know if there is a reason for it, that's why I included the scalar definition on both c an h files. 
But it works both ways.

garth-wells: The `.c` does not include the `.h` file to avoid mixing C/C++. The `.h` needs to be valid C/C++, while the `.c` file need to be C. 

C is not a strict subset of C++, so it is good to keep things separate as far as is possible.

blechta: Ok, so `.c` file is C implementation and `.h` file is a C/C++ layer on top of it?

garth-wells: More-or-less.

It is also for `cffi` that I made the split. There is an issue with includes and `cffi`. Let's avoid including the `.h` file in the `.c` file for as long as we can.

PR #35, opened 2018-06-25
Declare sub- elements and dofmaps

chrisrichardson:
Missing C declarations for sub_dofmaps and sub_elements cause some C99 compliant compilers to assume `int` return type instead of pointer. This usually causes a SEGV. 

This PR adds C declarations for sub_dofmaps and sub_elements.


chrisrichardson:
Can't be in `ufc.h` - it is the declaration of `create_foo_hash()` from another element.

PR #37, opened 2018-07-06
Cross-element code generation in UFLACS

w1th0utnam3:
This PR implements basic cross-element `tabulate_tensor` code generation in UFLACS. Before merging, the following points are still open:

 - Implement some unit tests comparing computed values of cross-element and default `tabulate_tensor` functions?
 - Return a different signature for `tabulate_tensor` if vector extensions are enabled (requires `double4` input instead of `double`) - however, this happens in a different stage than the actual code generation
 - Disable `enable_cross_element_gcc_ext` if the form uses complex variables as vector extensions don't work with the native complex types (should work in the "simple-loop" mode without vector extensions, untested)
 - Better names for the added FFC parameters?
 - Move the `vectorize_` functions from `integralgenerator.py` to a separate file? Maybe to the optimization stage?
 - Either remove the additional "optimization" flags in FFC parameters such as `enable_cross_element_fuse` and `enable_cross_element_array_conv` because they generally don't improve performance or implement them such that they are also supported in `enable_cross_element_gcc_ext` mode, didn't test them yet with vector extensions

Furthermore, cross-element assembly is currently not supported by DOLFIN but this is a separate issue from the code generation.


w1th0utnam3:
Yes, I think in general it would make sense because it's more an optimization step than any of the stuff that is currently in the class. However, this requires thinking a little bit about what data this new class requires from the integral generator. As of now, I think I only use the "CNodes language instance" and the type of the integral but no detailed information.

w1th0utnam3:
Will be replaced with a new PR from other branch.

PR #43, opened 2018-07-10
Dead code removal

w1th0utnam3:
As far as I can see the following code is dead and the suggested functionality implemented through other functions, e.g. `ffc/backends/ufc/integrals.py`:  `enabled_coefficients(self, L, ir)` is replaced by `ffc/representation/utils.py`: `generate_enabled_coefficients(enabled_coefficients)`

The integrals metadata parameter `num_cells` (intended for custom integrals?) isn't used at all.


garth-wells:
@chrisrichardson does this follow from your enabled coefficients changes?

chrisrichardson:
I don't think I did anything with this.

PR #44, opened 2018-07-11
Simple test framework based on cffi

garth-wells:
Add some support for jitting objects using cffi, and using to writ unit tests.

No doubt there is a lot more to add and improvement to the framework, but adding now allows tests to be added. I started this a while back, and it would in fact have already picked some issues had I merged it sooner.


blechta:
BTW, why UFLACS unit tests had to go?

garth-wells:
@blechta I don't know what you mean by 'had to go', but the uflacs test were not being run. Someone can of course add back the ones that run, and clean them up (they remain in the history). The current test files are all run, which is better in my view than having a bunch if non-functioning tests lying about.

PR #47, opened 2018-07-13
Add scalar_type to command line options

IgorBaratta:
The command 
`python3 -m ffc -l dolfin -fscalar_type='double complex' xxxx.ufl`
now throws an AssertionError (`assert key in parameters`).

With the addition of `scalar_type` to command line options, the following comannd
`python3 -m ffc -l dolfin -s 'double_complex' -r 'tsfc' xxxx.ufl`
works.

### DIFF
--- a/ffc/main.py
+++ b/ffc/main.py
@@ -63,7 +63,7 @@ parser.add_argument(
     action='store',
     choices=('double', 'double complex'),
     default="double",
-    help="Scalar type (default: double)")
+    help="scalar type to use (default: double)")
 parser.add_argument(
     '-f',
     action="append",
--- a/ffc/main.py
+++ b/ffc/main.py
@@ -57,6 +57,14 @@ parser.add_argument(
     default="uflacs",
     help="backend to use for compiling forms (default: uflacs)")
 parser.add_argument(
+    "-s",
+    "--scalar_type",
+    type=str,
+    action='store',
+    choices=('double', 'double complex'),
+    default="double",
+    help="Scalar type (default: double)")
+parser.add_argument(
     '-f',
     action="append",
     default=[],
@@ -91,6 +99,7 @@ def main(args=None):
     parameters["representation"] = xargs.representation
     parameters["quadrature_rule"] = xargs.quadrature_rule
     parameters["quadrature_degree"] = xargs.quadrature_degree
+    parameters["scalar_type"] = xargs.scalar_type
     if xargs.output_directory:
         parameters["output_dir"] = xargs.output_directory
     for p in xargs.f:


blechta:
> With the addition of scalar_type to command line options, the following comannd
python3 -m ffc -l dolfin -s 'double_complex' -r 'tsfc' xxxx.ufl
works.

I guess you mean `double complex`?

blechta:
Shouldn't this be handled by `-fscalar_type=...` to prevent bloating command-line interface by every possible option? @garth-wells

IgorBaratta:
> I guess you mean double complex?

Yes, thanks.
I fixed it on the description.

IgorBaratta:
> Shouldn't this be handled by -fscalar_type=... to prevent bloating command-line interface by every possible option? @garth-wells

I thought about that. But the way it's implemented, `scalar_type` would have to be added to the default parameters (not that this is a problem).

garth-wells:
I think for complex we should have an explicit option (we then get doc and with `ffc -h` and we can specify allowable values.

Should this be split into two options - one for the real/complex, and one for the float precision?

garth-wells:
Should complex type be inferred from the UFL code?

IgorBaratta:
I think it is a good idea to split into scalar_type and float precision.
Should it be something like: `scalar_type {real or complex}` and `float_precision{single,double, quad}`?

> Should complex type be inferred from the UFL code?
I guess one may want to generate complex code from a real UFL form.
and UFL itself (complex branch)  only knows that it's dealing with complex numbers in `compute_form_data`  because of a `complex_mode` parameter.

blechta:
As Igor says, UFL form does not know whether it is complex. It is a responsibility of form compiler's client (DOLFIN or user on command-line) to ask for real or complex.

I don't think that going into a split precision/complex specification is wise. There is already a meaning of `scalar_type` parameter (and its possible values)  as defined in TSFC which we adopted. Introducing another paramters system (now 2 paramters!) makes it more error-prone. If anything I would prefer to use NumPy dtype for the description:
```
>>> numpy.dtype('d')
dtype('float64')
>>> numpy.dtype('D')
dtype('complex128')
>>> numpy.dtype('double')
dtype('float64')
>>> numpy.dtype('float')
dtype('float64')
>>> numpy.dtype('float64')
dtype('float64')
>>> numpy.dtype('complex')
dtype('complex128')
>>> numpy.dtype('complex128')
dtype('complex128')
```
Let's not reinvent the wheel. Doc: https://docs.scipy.org/doc/numpy-1.14.5/reference/generated/numpy.dtype.html

blechta:
I would suggest to hold on this. It will possible change in TSFC.  In the mean time I would just recommend to add `scalar_type` to the list of parameters.

IgorBaratta:
Ok, I agree.
Also, when [PR #48](https://github.com/FEniCS/ffcx/pull/48) is merged, we could use:
`ffc -u scalar_type 'double complex' foo.ufl`.

garth-wells:
It is much tidier/simpler to process an option with a fixed number of arguments, e.g. 'complex64' or 'float64' rather than 'double complex' or 'double' (or is that 'complex double' . . . .)

blechta:
I would suggest to add `-z` short hand if needed.

PR #48, opened 2018-07-13
Support setting new and existing FFC parameter options.

garth-wells:
Allow setting of existing parameters:
```
ffc -f pname value foo.ufl
```
and adding new parameters:
```
ffc -u new_pname value foo.ufl
```
The type will always be `str`, so any casting needs to be handled internally.

### DIFF
--- a/ffc/main.py
+++ b/ffc/main.py
@@ -45,7 +45,7 @@ parser.add_argument(
     "--quadrature-rule",
     type=str,
     default="auto",
-    help="quadrature rule to apply (default: auto)")
+    help="quadrature rule to apply (default: %(default)s)")
 parser.add_argument(
     "--quadrature-degree",
     type=int,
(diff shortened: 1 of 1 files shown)


ffc/main.py
```diff
@@ -94,10 +106,15 @@ def main(args=None):
...
-        assert key in parameters
-        parameters[key] = value
+        assert len(p) == 2
+        if p[0] not in parameters:
+            raise RuntimeError("Command parameter set with -f does not exist in parameters system.")
+        parameters[p[0]] = p[1]
+    for p in xargs.u:
+        assert len(p) == 2
+        if p[0] in parameters:
+            raise RuntimeError("Command parameter set with -u alreadt exists in parameters system. Use -f.")
```
blechta: `alreadt` -> `already`

PR #57, opened 2018-09-07
Change Bessel functions to use C (with XSI extension if available)

chrisrichardson:


miklos1:
Oddly enough, we just had a user complain about this not working in complex.

PR #61, opened 2018-09-14
Cleanup analysis

chrisrichardson:
* Remove ~30% dead code from `uflacs/analysis`.
* Produces same output as before, and is slightly faster, even on a complex model like Holzapfel.


chrisrichardson:
No, but that is a pain to do when the repositories are split. I can make a dolfin branch to test.

PR #75, opened 2018-11-20
Remove some instances of ufl MultiFunction

chrisrichardson:
Removing `ufl.corealg.MultiFunction` from various places in ffc. Multifunction operates by calling different methods based on the ufl type of the first argument. This can be effectively achieved using a simple `dict` instead.


chrisrichardson:
NB this PR is not really "failing", just the coverage went down slightly.

ffc/backends/ffc/access.py
```diff
@@ -31,10 +30,29 @@ def __init__(self, ir, language, symbols, parameters):
         self.symbols = symbols
         self.parameters = parameters
 
-    # === Rules for all modified terminal types ===
-
-    def expr(self, e, mt, tabledata, num_points):
-        raise FFCError("Missing handler for type {0}.".format(e._ufl_class_.__name__))
+        self.call_lookup = {ufl.coefficient.Coefficient: self.coefficient,
```
garth-wells: Could you add some comments on what `self.call_lookup` is?

PR #81, opened 2018-11-26
Straighten backends

michalhabera:
Rearrange files from `ffc/backends` into `ffc/codegeneration`.


ffc/codegeneration/__init__.py
```diff
@@ -1,13 +1,16 @@
 # -*- coding: utf-8 -*-
-"""Utility functions for UFC"""
-
-__author__ = "FEniCS Project"
-__version__ = "2018.2.0.dev0"
-__license__ = "This code is released into the public domain"
+# Copyright (C) 2011-2018 Michal Habera and Martin Sandve Alnæs
```
garth-wells: Need to be careful changing the license here.

michalhabera: Just copied from some other file and added my name. What should I do? I do not insist on my name to be there.

garth-wells: Adding name is fine if you've made changes. Issue is the license change which can't be done without permission of all copyright holders.

michalhabera: @garth-wells Fixed

PR #93, opened 2019-01-22
Add cffi builder for ufc dofmap and coordinate_mapping

chrisrichardson:
This is a first step towards eliminating dijitso from ffc/dolfinx.


ffc/codegeneration/jit.py
```diff
@@ -271,26 +255,48 @@ def compile_forms(forms, module_name=None, parameters=None):
     for name in form_names:
         decl += form_template.format(name=name)
 
-    _, code_body = ffc.compiler.compile_form(forms, parameters=parameters)
+    _, code_body = ffc.compiler.compile_form(forms, parameters=p)
+
+    return _compile_objects(decl, code_body, form_names, module_name, p)
+
+
+def compile_coordinate_maps(cmaps, module_name=None, parameters=None):
```
michalhabera: There is `compile_coordinate_maps ` and `compile_coordinate_mapping `. 

I am slightly worried about ambiguous meaning of `compile` everywhere. Once its compilation of C code done by cffi/dijitso, on the other hand its compilations of ufl objects performed by ffc (results in code string). That makes it difficult to follow, some renaming might be useful in future.

chrisrichardson: I agree, we need to do some name changes.
compile = .ufl -> .c
or compile = .c -> .so
Which one should we change, and what to?

michalhabera: compile is used within FFC extensively with meaning .ufl -> .c. Lets just keep it, its THE compilation in this context. Maybe prefix other compilations with the "compiler" in that case?
cffi_compile = .c -> .so

PR #95, opened 2019-01-23
Rewrite analysis

michalhabera:
Simplified analysis step, with NumPy style docs.

Some corner cases might be buggy, I'll add tests later.

Dolfinx tested here https://github.com/FEniCS/dolfinx/commits/michal/ffc-rewrite-analysis


chrisrichardson:
Would be good to test against dolfin.

PR #102, opened 2019-02-03
Use cffi as the only JIT compiler

chrisrichardson:
* removes dijitso dependency
* simplifies code generation
* uses a simple cache directory (user configurable), currently `compile_cache` in current dir.
* interfaces to dolfin, see branch: https://github.com/FEniCS/dolfinx/tree/chris/cffi-jit


.circleci/config.yml
```diff
@@ -32,3 +32,8 @@ jobs:
             export PATH=$PATH:$HOME/.local/bin
             coverage run --source ffc -m pytest -s test/
             coveralls
+      - run:
+          name: Rerun tests (check caching)
+          command: |
+            export PATH=$PATH:$HOME/.local/bin
```
garth-wells: Why is the path addition required?

ffc/classname.py
```diff
@@ -6,6 +6,11 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 """Some basics for generating C code."""
 
+import ffc
+import ufl
+import hashlib
+from ffc.parameters import compute_jit_parameters_signature
```
garth-wells: Could `compute_jit_parameters_signature` be just `compute_jit_signature` since it's already namespaced by the module  `parameters`?

ffc/classname.py
```diff
@@ -16,3 +21,50 @@ def make_name(prefix, basename, signature):
 def make_integral_name(prefix, integral_type, form_id, subdomain_id):
     basename = "{}_integral_{}".format(integral_type, str(form_id).lower())
     return make_name(prefix, basename, subdomain_id)
+
+
+def compute_signature(ufl_objects, tag, parameters, coordinate_mapping=False):
+    """Compute the signature hash for jit modules.
```
garth-wells: Maybe comment on what precisely is being hashed.

ffc/codegeneration/jit.py
```diff
@@ -214,86 +217,162 @@
 """
 
 
+def get_cached_module(module_name, object_names, parameters):
+    cache_dir = parameters.get("cache_dir",
+                               "compile_cache")
+
+    timeout = int(parameters.get("timeout", 100))
+
+    c_filename = cache_dir + "/" + module_name + ".c"
```
garth-wells: Use `pathlib` (https://docs.python.org/3/library/pathlib.html#module-pathlib) to manipulate paths and filenames.

ffc/codegeneration/jit.py
```diff
@@ -214,86 +217,162 @@
...
 
-    cache_dir = None
-    if parameters:
-        cache_dir = parameters.get("cache_dir")
-    if not cache_dir:
-        cache_dir = "compile_cache"
+    cache_dir = parameters.get("cache_dir",
+                               "compile_cache")
 
+    c_filename = cache_dir + "/" + module_name + ".c"
```
garth-wells: Use `pathlib`.

ffc/parameters.py
```diff
@@ -52,7 +55,7 @@
     "external_include_dirs": "",
 }
 _FFC_CACHE_PARAMETERS = {
-    "cache_dir": "",  # cache dir used by Instant
+    "cache_dir": "compile_cache",  # cache dir used by default
```
garth-wells: Use `pathlib` to make clear that the strings are actually filesystem directory names.

setup.py
```diff
@@ -21,7 +21,6 @@
         "numpy",
         "fenics-fiat{}".format(RESTRICT_REQUIREMENTS),
         "fenics-ufl{}".format(RESTRICT_REQUIREMENTS),
-        "fenics-dijitso{}".format(RESTRICT_REQUIREMENTS),
     ]
```
garth-wells: Does cffi need to be added as  a dependency?

.circleci/config.yml
chrisrichardson: Don't know - I just copied from previous invocation above.

ffc/classname.py
chrisrichardson: 👍

.circleci/config.yml
garth-wells: Try removing - shouldn't be needed in this case. The case above calls some binaries in `$HOME/.local/bin` but this case doesn't.

PR #112, opened 2019-03-24
Set new ufl flag

chrisrichardson:


ffc/analysis.py
```diff
@@ -171,6 +171,7 @@ def _analyze_form(form: ufl.form.Form, parameters: Dict) -> ufl.algorithms.formd
             do_apply_geometry_lowering=True,
             preserve_geometry_types=(ufl.classes.Jacobian, ),
             do_apply_restrictions=True,
+            do_append_everywhere_integrals=False,
```
garth-wells: Maybe add a comment on what the new option is for.

PR #117, opened 2019-03-26
Remove crosslinking 

garth-wells:
Removes linking between libraries.

Complicated matter for only small gain. Can be added later (and make simpler) if needed.


garth-wells:
Yes, tested against DOLFIN.

PR #121, opened 2019-04-02
Rename ir to r in comprehensions to aid code readability

chrisrichardson:


garth-wells:
For me, using `r` reduces readability. More verbose `ir_dofmap`, etc, would be better.

chrisrichardson:
Sure, I'll change them to something better.

PR #136, opened 2019-06-18
Improve jit code

chrisrichardson:
* Get the declarations for JIT directly from `ufc.h`
* Add a parameter `use_cache` (default = True) 
  - when set to false, modules are compiled in a temp directory

### DIFF
--- a/ffc/codegeneration/jit.py
+++ b/ffc/codegeneration/jit.py
@@ -10,6 +10,7 @@ import logging
 import os
 import sys
 import time
+import re
 
 import cffi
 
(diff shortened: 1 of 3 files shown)


ffc/codegeneration/jit.py
```diff
@@ -298,9 +164,10 @@ def compile_coordinate_maps(meshes, parameters=None):
     cmap_names = [ffc.ir.representation.make_coordinate_mapping_jit_classname(
         mesh.ufl_coordinate_element(), "JIT", p) for mesh in meshes]
 
-    obj, mod = get_cached_module(module_name, cmap_names, p)
-    if obj is not None:
-        return obj, mod
+    if p['use_cache']:
```
garth-wells: This doesn't really make sense when `p` has a default value of `None`. Will lead to runtime error.

The are other examples of this issue above.

chrisrichardson: it doesn't crash because of the call to `validate_parameters()`, something I haven't changed. Not very clear code, I agree.

garth-wells: Ok. I find `validate_parameters()` unclear, and in fact misleading. It doesn't just validate - it changes the parameter `dict`.

chrisrichardson: Agree. It was there from before. I can make a change and add it to this PR.

PR #138, opened 2019-06-26
Revise JIT module import

garth-wells:
This addresses an issue with Python module import caching which can lead to JIT-created modules being overlooked due to Python not re-checking module search paths.

### DIFF
--- a/ffc/codegeneration/jit.py
+++ b/ffc/codegeneration/jit.py
@@ -57,7 +57,7 @@ def get_cached_module(module_name, object_names, parameters):
     """Look for an existing C file and wait for compilation, or if it does not exist, create it."""
     cache_dir = ffc.config.get_cache_path(parameters)
     timeout = int(parameters.get("timeout", 10))
-    c_filename = cache_dir.joinpath(module_name + ".c")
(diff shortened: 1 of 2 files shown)


chrisrichardson:
Import happens in two places (direct after compile, or from cache) so need to put something in both places, I think.

ffc/codegeneration/jit.py
```diff
@@ -195,27 +200,35 @@ def _compile_objects(decl, ufl_objects, object_names, module_name, parameters):
...
+    # (new) modules are found
+    finder.invalidate_caches()
+    spec = finder.find_spec(module_name)
+    if spec is None:
+        raise ModuleNotFoundError("Unable to find JIT module.")
+
+    # Load module
+    compiled_module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(compiled_module)
+
```
chrisrichardson: Since this is repeated code, I wonder if it would be better in a small function, e.g. `def _load_module(name, path)`

garth-wells: I'd rather keep it as-is for readability and clarity. Also, I would like to revise the function `get_cached_module`.

PR #142, opened 2019-08-16
Add "add" mode

chrisrichardson:
* Adds a new `tensor_init_mode` called `add` for testing (simple tests in PR)
* Does not zero the tensor before filling
* Accumulates values
* Should become the default in the future


garth-wells:
@chrisrichardson Any reason to make this an option rather than just the normal behaviour?

We should minimise options for simplicity and for the massive reduction in testing complexity.

PR #144, opened 2019-08-25
Add new Constant support

michalhabera:
Based on https://bitbucket.org/fenics-project/ufl/branch/michal/ufl-constant this implements the support for form constants which are not FE based.

New constants are passed to tabulation kernels in additional argument `tabulate_foo(..., const PetscScalar* c, ...)` which allows separation from other FE-based passed data (coefficients `w`).
Additional argument could be in future used for any other non-FE data which has to be passed to the kernel.

### DIFF
--- a/ffc/codegeneration/access.py
+++ b/ffc/codegeneration/access.py
@@ -51,11 +51,19 @@ class FFCBackendAccess(object):
 
     def get(self, e, mt, tabledata, num_points):
         # Call appropriate handler, depending on the type of e
-        etype = type(e)
-        if etype in self.call_lookup:
-            return self.call_lookup[etype](e, mt, tabledata, num_points)
+        handler = self.call_lookup.get(type(e), False)
+
+        if not handler:
+            # Look for parent class types instead
+            for k in self.call_lookup.keys():
+                if isinstance(e, k):
+                    handler = self.call_lookup[k]
+                    break
+
+        if handler:
+            return handler(e, mt, tabledata, num_points)
         else:
-            raise RuntimeError("Not handled: %s", etype)
+            raise RuntimeError("Not handled: %s", type(e))
 
     def coefficient(self, e, mt, tabledata, num_points):
         ttype = tabledata.ttype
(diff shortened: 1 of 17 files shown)


ffc/ir/representation.py
```diff
@@ -390,10 +391,14 @@ def _compute_form_ir(form_data, form_id, prefix, element_numbers,
 
     ir["rank"] = len(form_data.original_form.arguments())
     ir["num_coefficients"] = len(form_data.reduced_coefficients)
+    ir["num_constants"] = len(form_data.original_form.constants())
 
     ir["coefficient_names"] = [object_names.get(id(obj), "w%d" % j)
                                for j, obj in enumerate(form_data.reduced_coefficients)]
 
+    ir["constant_names"] = [object_names.get(id(obj), "w%d" % j)
```
chrisrichardson: Should this be `c%d` ?

ffc/codegeneration/access.py
```diff
@@ -50,11 +51,19 @@ def __init__(self, ir, language, symbols, parameters):
...
+            for k in self.call_lookup.keys():
+                if isinstance(e, k):
+                    handler = self.call_lookup[k]
+                    break
+
+        if handler:
+            return handler(e, mt, tabledata, num_points)
         else:
-            raise RuntimeError("Not handled: %s", etype)
+            raise RuntimeError("Not handled: %s", type(e))
```
garth-wells: Use Python3-style string formatting.

ffc/codegeneration/integrals_template.py
```diff
@@ -16,6 +16,7 @@
     "cell":
     """
 void tabulate_tensor_{factory_name}(ufc_scalar_t* restrict A, const ufc_scalar_t* w,
+                                    const ufc_scalar_t* c,
```
garth-wells: Not specific to this PR, but can we reduce code duplication now that all tabulate function have a common signature?

chrisrichardson: It's not that easy, because the fourth argument `entity_local_index` is named differently in the generated code, as `facet`, `vertex` etc. depending on which version is being called. It doesn't matter for the signature, but they are different for the implementation. It is worth considering how to address in another PR, agreed.

Issue #145, opened 2019-08-27
Max/min of geometric quantities


michalhabera:
In complex mode, `CellDiameter` expands to maximum over cell edges lengths (this happens in UFL).
MaxValue has no handler in `ufl_to_cnodes.py` which is safe, because maximum of complex number is ambiguous.
Since geometry in FFCx/dolfinx is always real, we should support MaxValue and MinValue on geometric quantities.

michalhabera:
Resolved with #146 .

PR #146, opened 2019-08-28
Handle complex max/min

michalhabera:
Addresses https://github.com/FEniCS/ffcx/issues/145

The complex case then falls to `fmax/fmaxf` which basically compares only first `double/float` bits of the complex number, effectively taking only maximum of the real parts of the numbers.

This is safe when arguments are complex but with zero imaginary part (geometric quantities). However, it does not prevent user from `Max(Coefficient())`.

PETSc VecMax/Min takes the maximum of real part only too.


chrisrichardson:
It seems a bit unsatisfactory. If you want to use `CellDiameter` and it uses `MaxValue` in complex mode, it will issue warnings (which are not needed, since the geometry is not complex).

PR #153, opened 2019-09-27
Turn on compiler auto vectorisation

michalhabera:
What is JIT compiled is always used on the target machine, we should enable native simd intrinsics.


mrambausek:
Is this assumption true for clusters? I am thinking of a cluster that features some "old" and some 
brand-new/replaced nodes... Can this be resolved by using a "node-local" jit cache?

michalhabera:
JIT compilation happens on a root process (rank == 0). It would indeed cause a problem if the root process has a different architecture than other processes. However, I think we should not penalise 99.9% of users for this rare heterogeneous cluster case. More advanced user can force environmental `CFLAGS` and be more conservative.

mrambausek:
@michalhabera Thanks a lot -- the CFLAGS thing is all I need to know.

PR #154, opened 2019-09-30
Fix offsets calculation for interior facet integrals

michalhabera:
This PR changes layout of coefficients for form.
New layout is flattened `w[coefficient][restriction][dof]`.

For `f * ufl.dx + g * ufl.dS` FFC generates two integral tabulation kernels. Cell integral expects coefficients as
`[f_dof0, f_dof1, f_dof2, g_dof0, g_dof1, g_dof2]`
while interior facet integral expects
`[f_r0_dof0, f_r0_dof1, f_r0_dof2, f_r1_dof0, f_r1_dof1, f_r1_dof2, ... the same pattern for g]`.

### DIFF
--- a/ffc/ir/uflacs/uflacsrepresentation.py
+++ b/ffc/ir/uflacs/uflacsrepresentation.py
@@ -106,10 +106,17 @@ def compute_integral_ir(itg_data, form_data, form_id, element_numbers, classname
 
     index_to_coeff = sorted([(v, k) for k, v in coefficient_numbering.items()])
     offsets = {}
+
+    if integral_type in ("interior_facet"):
+        # For interior facet integrals we need + and - restrictions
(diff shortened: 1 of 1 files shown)


garth-wells:
Is there a test we could add to FFC that would pick this issue up?

michalhabera:
It is not an issue from FFCx perspective. The old layout produced a working code. This is to match the layout that dolfinx assembler expects.

For the form above the old layout for interior facet integral expected in FFC was `[f_dof0, f_dof1, f_dof2, g_r0_dof0, g_r0_dof1, g_r0_dof2, g_r1_dof0, g_r1_dof1, g_r1_dof2]` -- it effectively expected restriction only for the coefficients that are active in interior facet integral.

PR #158, opened 2019-10-08
Dokken/fixissue157

jorgensd:
Removing the local hack for finding number of dofs for a coordinate_element component, using fiat interface.

### DIFF
--- a/ffc/codegeneration/definitions.py
+++ b/ffc/codegeneration/definitions.py
@@ -18,8 +18,11 @@ def num_coordinate_component_dofs(coordinate_element):
 
     """
     from ffc.fiatinterface import create_element
+    fiat_elements = create_element(coordinate_element).elements()
     # Extracting only first component degrees of freedom from FIAT
-    fiat_element = create_element(coordinate_element).elements()[0]
+    fiat_element = fiat_elements[0]
+    assert(all(isinstance(element, type(fiat_element)) for element in fiat_elements))
+
     dof_dict = fiat_element.entity_dofs()
     d = 0
     for entity_dim in dof_dict.keys():
--- a/ffc/codegeneration/definitions.py
+++ b/ffc/codegeneration/definitions.py
@@ -22,13 +22,7 @@ def num_coordinate_component_dofs(coordinate_element):
     # Extracting only first component degrees of freedom from FIAT
     fiat_element = fiat_elements[0]
     assert(all(isinstance(element, type(fiat_element)) for element in fiat_elements))
-
-    dof_dict = fiat_element.entity_dofs()
-    d = 0
-    for entity_dim in dof_dict.keys():
-        for enitity in dof_dict[entity_dim]:
-            d += len(dof_dict[entity_dim][enitity])
-    return d
+    return fiat_element.space_dimension()
 
 
 class FFCBackendDefinitions(object):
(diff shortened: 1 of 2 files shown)


ffc/codegeneration/definitions.py
```diff
@@ -16,34 +16,15 @@
...
-    The table data is to messy to figure out a clean design for that quickly.
     """
-    from ufl.cell import num_cell_entities
-    degree = coordinate_element.degree()
-    cell = coordinate_element.cell()
-    tdim = cell.topological_dimension()
-    cellname = cell.cellname()
+    from ffc.fiatinterface import create_element
+    # Extracting only first component degrees of freedom from FIAT
+    fiat_element = create_element(coordinate_element).elements()[0]
```
michalhabera: Please assert that all components have the same element.

ffc/codegeneration/definitions.py
```diff
@@ -16,34 +16,18 @@
...
-            else:
-                n = (degree - 2) * (degree - 1) // 2
-        elif entity_dim == 3:
-            n = (degree - 3) * (degree - 2) * (degree - 1) // 6
-        else:
-            raise RuntimeError("Entity dimension out of range")
-        # Accumulate
-        num_entities = num_cell_entities[cellname][entity_dim]
-        d += num_entities * n
+    for entity_dim in dof_dict.keys():
```
michalhabera: What about using
```
    def space_dimension(self):
        """Return the dimension of the finite element space."""
        return len(self.dual_basis())
```
?

`entity_dofs` doesnt return e.g. dofs of DG elements, which live in the interior of the cell, you'd need to use `entity_closure_dofs`. I know coordinate map is not discontinuous, but just for clarity.

jorgensd: Sure, seems ok. Reason for the other version was to extend the old hack to arbitrary lagrangian elements.

michalhabera: One more comment. You can notice currently, that FIAT is interfaced only in `fiatiinterface.py`. Can you refactor a but and query FIAT library only in `fiatinterface.py`?

jorgensd: Is that true? I can see it accessed in: codegeneration/access.py

michalhabera: Ah, sorry. I thought you were importing FIAT library, its fine like this, nvm.

michalhabera:
@jorgensd Do you have a dolfinx branch to test against?

ffc/codegeneration/definitions.py
```diff
@@ -16,35 +16,13 @@
...
-                n = (degree - 2) * (degree - 1) // 2
-        elif entity_dim == 3:
-            n = (degree - 3) * (degree - 2) * (degree - 1) // 6
-        else:
-            raise RuntimeError("Entity dimension out of range")
-        # Accumulate
-        num_entities = num_cell_entities[cellname][entity_dim]
-        d += num_entities * n
-    return d
+    from ffc.fiatinterface import create_element
```
chrisrichardson: I thought flake8 was supposed to stop you importing in the middle of a file...?

jorgensd: Good question. Have moved it regardless

jorgensd:
@michalhabera https://github.com/FEniCS/dolfinx/tree/dokken/ffc-fixissue157

PR #159, opened 2019-10-14
[WIP] Rename UFC as FEniCS interface

michalhabera:
Renamed UFC as FEniCS interface. Flattened the tests directories.

UFC is a legacy name without much meaning for users. New design is that FFCx generates a code which complies to the FEniCS interface, this interface is temporarily included also in ffc in `fenics_interface.h` and `fenics_geometry.h`.


michalhabera:
Closing now, need to reimplement in the future due to many changes in the master.

PR #160, opened 2019-10-14
Add expression tabulation

michalhabera:
This add a possibility to generate a code for pointwise evaluation of UFL expressions.

FFCx expects an UFL expression attached with an array of points at which to evaluate. It generates a C routine for this, with signature very similar to the integral tabulation kernels, but appends dimensions of tabulated tensor with dimensions of tabulated expression (ufl's shape) and with number of points at which to evaluate.

How this is done:
1. Analysis step
    Expressions are preprocessed trivially, with lowering algebra expressions, applying derivatives and applying pullbacks.
2. Intermediate representation (IR) step
    An UFL expression is given a new custom integral type  == "expression". This is for the sake of reusing UFLACS IR. Expressions pretend to be an integral of new integral type which just needs evaluation at points in reference cell.
Since in general we need expression to be tensor-valued, unlike integrands which are usually just scalar valued, this needs a major work in how expression trees are built. New implementation allows tensor valued expression trees and their factorization (also with Arguments).

    First, scalar tree is built which has as many root nodes as there are components of the evaluated expression. This already introduces subexpression elimination across components, i.e. if the expression is symmetric the built scalar tree recognises it. All root (target) nodes store variable `comp` to denote which component they contribute to. 

    Later, the expression tree is factorised. Factorised tree has as many root nodes as there are identified unique factors or arguments. Additionally, these root nodes store also all components to which they contribute in `comp` list. Since there could be more components to which the factor contributes, there are more argument indices, too.

3. Code generation
    `IntegralGenerator` could not have been used because it is summing over quadrature points. For this reasons new `ExpressionGenerator` was added, which has some similarities with IntegralGenerator, so this allows refactoring to identify common concepts in future. I decided not to make expression generation just a special case of integral generation, since many specialisations and simplifications are possible for expression generation.


ffc/codegeneration/ufc.h
```diff
@@ -438,6 +438,15 @@ extern "C"
     ufc_tabulate_tensor_custom* tabulate_tensor;
   } ufc_custom_integral;
 
+  typedef struct ufc_expression
```
garth-wells: The reference coords at which the expression is being evaluated, the number of points and the return shape should be provided as members.

ffc/codegeneration/expressions.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
garth-wells: This line isn't required with Python 3.

ffc/codegeneration/expressions_template.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
garth-wells: Not required

ffc/codegeneration/expressions_template.py
```diff
@@ -0,0 +1,43 @@
...
+  {original_coefficient_positions}
+  {points}
+  {value_shape}
+
+  ufc_expression* expression = malloc(sizeof(*expression));
+  expression->tabulate_expression = tabulate_expression_{factory_name};
+  expression->original_coefficient_positions = original_coefficient_positions;
+  expression->num_coefficients = {num_coefficients};
+  expression->num_points = {num_points};
+  expression->point_dim = {point_dim};
```
garth-wells: Should this just be `dim`, or maybe `topological_dimension`?

ffc/ir/uflacs/analysis/factorization.py
```diff
@@ -299,39 +299,45 @@ def compute_argument_factorization(S, rank):
...
-            for argkey, fi in S.nodes[S_target]['factors'].items()
-        }
-    # Expecting all term keys to have length == rank
-    # (this assumption will eventually have to change if we
-    # implement joint bilinear+linear form factorization here)
-    assert all(len(k) == rank for k in factors)
+            # Forms of arity 1 or higher:
+            # Map argkeys from indices into SV to indices into AV,
+            # and resort keys for canonical representation
+            for argkey, fi in S.nodes[S_target]['factors'].items():
```
garth-wells: If find `fi` a bit difficult to parse. Is there are better name?

ffc/ir/uflacs/analysis/graph.py
```diff
@@ -73,14 +74,14 @@ def build_scalar_graph(expression):
     """Build list representation of expression graph covering the given expressions."""
 
     # Populate with vertices
-    G = build_graph_vertices(expression, scalar=False)
+    G = build_graph_vertices([expression], skip_terminal_modifiers=False)
 
-    # Build more fine grained computational graph of scalar subexpressions
-    scalar_expression = rebuild_with_scalar_subexpressions(G)
+    # # Build more fine grained computational graph of scalar subexpressions
```
garth-wells: Remove extra `#`.

test/ufc/test_expression.py
```diff
@@ -0,0 +1,74 @@
...
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+
+import numpy as np
+import cffi
+
+import ufl
+import ffc.codegeneration.jit
+
+
+def float_to_type(name):
```
garth-wells: Could you add more comments to the test to help someone understand how to use the new functionality?

ffc/ir/uflacs/analysis/factorization.py
michalhabera: `fi` is used on many places. There is a comment explaining the meaning. It is an index (in factorised graph) of current argument factor.

PR #161, opened 2019-10-15
Add assume aligned

chrisrichardson:
Just adds a few optional statements to `tabulate_tensor` to help the compiler with vectorisation.

* Use e.g. `ffc -u assume_aligned 64 Foo.ufl`
* Works with clang, gcc and intel.
* Requires caller to use aligned memory (obviously)


michalhabera:
Does it really work for intel and clang, or just doesn't hurt?

jhale:
Worth remembering the extensive report by @w1th0utnam3 on this topic: https://gist.github.com/w1th0utnam3/6ca59f112572dc166048be78d9e631c5

chrisrichardson:
@michalhabera - I'm not sure about clang, but it seems to have a real effect on intel. 
@jhale - thanks, I had forgotten about it. This is a very simple change by comparison. The analysis of Eigen alignment in Fabian's article is useful to reread.

michalhabera:
Weird, intel seems to be using `__assume_aligned` and clang `__attribute__((assume_aligned(alignment)))`, see http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0886r0.pdf

We should check with godbolt.

garth-wells:
I'd suggest we need to do this via an `DEFINE` or something so the code that calls the functions can determine the alignment.

chrisrichardson:
@garth-wells - yes, that would be good. I think it is a minimal change - we can just make it a string parameter instead of an int, so it can be: `ffc -u assume_aligned EIGEN_DEFAULT_ALIGN_BYTES Foo.ufl` for example.

PR #173, opened 2019-10-30
Changing FFCx ordering to be TensorProduct for quads and hexes to match FIAT

jorgensd:
Fixes bug in Quadrilateral meshes.
For second order meshes to be correct, the order in `basisvalues` should be lexicographic, with
x=0, y=0,..,p-1
x=1, y=-,...,p-1

Added test to check that the layout of a cell should be FIAT TensorProduct, i.e. 
```
1--7--4
|        |
2  8   5
|        |
0--6--3
```

### DIFF
--- a/test/ufc/test_cffi.py
+++ b/test/ufc/test_cffi.py
@@ -126,19 +126,18 @@ def test_cmap_quads(degree, coords):
     e = ufl.VectorElement("Lagrange", cell, degree)
     mesh = ufl.Mesh(e)
     compiled_cmap, module = ffc.codegeneration.jit.compile_coordinate_maps([mesh])
-    X = np.zeros_like(x)
-    X_ptr = module.ffi.cast("double *", module.ffi.from_buffer(X))
 
     coords_ptr = module.ffi.cast("double *", module.ffi.from_buffer(coords))
 
     x = np.array([[1 / 3, 1 / 3]], dtype=np.float64)
     x_ptr = module.ffi.cast("double *", module.ffi.from_buffer(x))
-    X_physical = np.zeros_like(x)
-    X_physical_ptr = module.ffi.cast("double *", module.ffi.from_buffer(X_physical))
-    compiled_cmap[0].compute_physical_coordinates(X_physical_ptr, X.shape[0], x_ptr, coords_ptr)
+    X = np.zeros_like(x)
+    X_ptr = module.ffi.cast("double *", module.ffi.from_buffer(X))
 
-    assert(np.isclose(X_physical[0, 0], 3 * x[0, 0]))
-    assert(np.isclose(X_physical[0, 1], 2 * x[0, 1]))
+    compiled_cmap[0].compute_physical_coordinates(X_ptr, X.shape[0], x_ptr, coords_ptr)
+
+    assert(np.isclose(X[0, 0], 3 * x[0, 0]))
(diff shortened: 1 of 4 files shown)


michalhabera:
Is there an associated quad test in dolfinx that you test against? Can you check if the same bug is not in hexahedral cell?

jorgensd:
> Is there an associated quad test in dolfinx that you test against? Can you check if the same bug is not in hexahedral cell?

I've got this branch: https://github.com/FEniCS/dolfinx/tree/dokken/mesh-input-structure
which tests higher order quads, as well as changing the default input structure in dolfin to FEniCS ordering (UFC for simplices, TP for non-simplices). Just waiting for https://github.com/FEniCS/dolfinx/pull/586 to go through, so that we can  make even better tests. 

I'm gonna have a look at hexahedrons today.

garth-wells:
It would be nice to get a test for hexes into this PR.

jorgensd:
> It would be nice to get a test for hexes into this PR.

The problem with hexes is that the first order tests are not sensitive to the Lexicographic vs. TP ordering. This only occurs for second order hexes. 
I've started implementing them in FEniCSx, but hit a bump in the road due to some bugs in Paraview wrt, visualization:
https://gitlab.kitware.com/paraview/paraview/issues/19433
I can make a temporary work around (A custom permutation as I've figured out what it should be). 
However, this will make the dolfin PR quite bug.

garth-wells:
@jorgensd Can tests for hexes not be added to this PR (FFC)? The is a `test_cmap_quads` test.

jorgensd:
> @jorgensd Can tests for hexes not be added to this PR (FFC)? The is a `test_cmap_quads` test.

I have a failing test for it locally. I can add it (and fix the issue). However, I would like to avoid added a lot of tests in the dolfinx layer.

jorgensd:
@michalhabera @garth-wells Second order hexes tested and fixed in latest commit

jorgensd:
@garth-wells Could you merge this?

test/ufc/test_cffi.py
```diff
@@ -113,6 +113,28 @@ def test_cmap():
...
+    e = ufl.VectorElement("Lagrange", cell, degree)
+    mesh = ufl.Mesh(e)
+    compiled_cmap, module = ffc.codegeneration.jit.compile_coordinate_maps([mesh])
+    x = np.array([[1, 0.5]], dtype=np.float64)
+    x_ptr = module.ffi.cast("double *", module.ffi.from_buffer(x))
+    X = np.zeros_like(x)
+    X_ptr = module.ffi.cast("double *", module.ffi.from_buffer(X))
+
+    coords_ptr = module.ffi.cast("double *", module.ffi.from_buffer(coords))
+    compiled_cmap[0].compute_reference_coordinates(X_ptr, X.shape[0], x_ptr, coords_ptr, 0)
```
garth-wells: Would using `compute_physical_coordinates` be a better test? The forward map is simpler.

jorgensd: Now i've added tests for both.

garth-wells:
Let's remove calls to `compute_reference_coordinates` in the test. I don't think it's needed, and it brings in other issues, like the use of Newton's methods and associated tolerances for mapping from physical to reference coords. The push-forward is a much simpler operations.

PR #177, opened 2019-11-07
Catch failed compile

chrisrichardson:
@nate-sime - can you confirm this is a useful patch? It should stop some annoying timeouts.


ffc/codegeneration/jit.py
```diff
@@ -265,7 +265,12 @@ def _compile_objects(decl, ufl_objects, object_names, module_name, parameters, c
 
     # Compile (ensuring that compile dir exists)
     cache_dir.mkdir(exist_ok=True, parents=True)
-    ffibuilder.compile(tmpdir=cache_dir, verbose=cffi_verbose, debug=cffi_debug)
+
+    try:
+        ffibuilder.compile(tmpdir=cache_dir, verbose=cffi_verbose, debug=cffi_debug)
+    except Exception:
```
nate-sime: Should this be a `TimeoutError`?

PR #179, opened 2019-12-13
Tabulate subelements only

michalhabera:
Fixes https://github.com/FEniCS/dolfinx/issues/684

Instead of tabulating the whole MixedElement and slicing the actual needed component this tabulates only sub elements needed for the actual component.

Dolfinx tested here https://github.com/FEniCS/dolfinx/tree/michal/tabulate-subcomponents-only

PR #181, opened 2019-12-18
Fix CI failure after recent updates

IgorBaratta:
After updates of _pydocstyle_ and _coveralls_, master is failing.

**_pydocstyle_:**
When the last agurment of a function is labeled `parameters`, _pydocstyle_ checking breaks. Parameter is a keyword in the numpy doc style, so the confusion. See failing tests in https://circleci.com/gh/FEniCS/ffcx/1079
An alternative is to change the argument  name to `params` for avoiding this failure.

**_coveralls_:**
_Coveralls_ requires `coverage<5.0`, but the docker container has `coverage=5`.
Error message:

> ERROR: coveralls 1.9.2 has requirement coverage<5.0,>=3.6, but you'll have coverage 5.0 which is incompatible

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
           name: Install dependencies
           command: |
             sudo apt-get update && sudo apt-get install --no-install-recommends graphviz libgraphviz-dev
-            sudo pip install cffi flake8 pydocstyle pytest pytest-cov pytest-xdist pygraphviz coveralls --upgrade
+            sudo pip install cffi flake8 pydocstyle pytest pytest-cov pytest-xdist pygraphviz 'coveralls<1.9' --upgrade
       - run:
           name: Install FEniCS dependencies
           command: |
@@ -30,3 +30,4 @@ jobs:
           command: |
             export PATH=$PATH:$HOME/.local/bin
             python -m pytest --cov=ffc/ -n 4 -v test/
+            coveralls
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/python:3.7
+      - image: circleci/python:3.8
     working_directory: ~/ffc-test
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/python:3.8
+      - image: circleci/python:3.7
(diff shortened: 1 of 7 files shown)


IgorBaratta:
> Why "Attributes"? This doesn't look right.

Indeed that's not right, only module level parameters should be named Attributes.
The _correct_ alternative is `args`.

IgorBaratta:
I renamed the argument `parameters` to `params`, in the cases that fail. 
Also I updated the PR description.

BTW, the style of the Python docstrings is not consistent. I can work on this in a new PR, if we agree in a specific style.

michalhabera:
@IgorBaratta I think the issue is that `D417` requires all arguments be described once you've decided for the lengthy docstring. I do not think this has to do with `parameters` keyword.

I have added skipping of D417 in my PR https://github.com/FEniCS/ffcx/pull/179/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R22

I would suggest to just fix the `coveralls` issue in this PR.

chrisrichardson:
I don't mind, but maybe Michal is right - you have got `parameters` in uflacsrepresentation.py and it seems to not be causing a problem.

IgorBaratta:
See for example, the  analysis.py case:
```
def analyze_ufl_objects(ufl_objects: typing.Union[typing.List[ufl.form.Form], typing.List[ufl.FiniteElement],
                                                  typing.List],
                        parameters: typing.Dict) -> ufl_data:
    """Analyze ufl object(s).

    Parameters
    ----------
    ufl_objects
        Objects to be analysed.
    parameters
        Dict of FFC parameters.

```

If we remove the description of parameters we have indeed D417, but if we add the description the error is:

- D410: Missing blank line after section ('Parameters')
- D417: Missing argument descriptions in the docstring (argument(s) parameters are missing descriptions in 'analyze_ufl_objects' docstring)
- D405: Section name should be properly capitalized ('Parameters', not 'parameters')
- D411: Missing blank line before section ('Parameters')
- D407: Missing dashed underline after section ('Parameters')
- D417: Missing argument descriptions in the docstring (argument(s) parameters, ufl_objects are missing descriptions in 'analyze_ufl_objects' docstring)

michalhabera:
We will see later. We can report as a bug to `pydocstyle` if argument named "parameters" turns out to be breaking something.

My point why D417 is not really nice is that you do not want to be forced to document arguments whose name is self-explanatory, if e.g. `analyze_ufl_objects` takes argument `ufl_objects` - describing it with docstring as "UFL objects to be analysed" is redundant.

michalhabera:
Can we close now @IgorBaratta ?

PR #182, opened 2019-12-19
Provide information of dof type in generated code

IgorBaratta:
This PR provides a simple way to get information about dof type (if dofs are simple point evaluations) from generated code. 
It is used in branch [igor/interpolation](https://github.com/FEniCS/dolfinx/tree/igor/interpolation) to determine if the element uses only simple point functionals for its degrees of freedom, and if not it throws a non-implemented error.


chrisrichardson:
I think this is fine just sending a `bool`. I'm against sending a list of strings - they should at least be evals instead, and then we would need to also pass the length of the array as another variable.

michalhabera:
@chrisrichardson We need to provide more information, not just if **all** are point evals. What about component point evaluation? Dolfin (whatever else) can decide what to do.

garth-wells:
We should provide as much information as is available. How about an enum?

IgorBaratta:
> We should provide as much information as is available. How about an enum?

A list of enums? Every dof may have a different type.

IgorBaratta:
What would be more effective: having a list or a set with type_tags/enum (unique tag identifiers)? 
The information of an individual dof would ever be used by dolfin?

For example, `element = ufl.FiniteElement("N1curl", cell, 3)` has 45 degrees of freedom, and only three unique type tags `{'IntegralMoment', 'PointEdgeTangent', 'PointFaceTangent'}`.

garth-wells:
Is it that each cell entity type has a particular dof type?

chrisrichardson:
I guess in the most general case, each dof could be a different type, but if using a list of enums (int), this needn't be too huge.

mscroggs:
For deciding which dofs correspond to vectors and need to be reversed, I need information about the dof type. I've added a version of this this in https://github.com/FEniCS/ffcx/commit/a012c72e8370b808578b47881c88fd2c4b051ffb

https://github.com/FEniCS/ffcx/blob/a012c72e8370b808578b47881c88fd2c4b051ffb/ffc/ir/representation.py#L203

Currently its a list of strings as this was easiest to implement.

IgorBaratta:
On hold until  #190 is merged.
It may no longer be necessary, or may change the scope.

PR #187, opened 2020-01-06
Fix vector elements

michalhabera:
Fixes issue in tabulation of vector elements which have 0 sub elements - i.e. vector/tensor valued basis functions (BDM, RT, Regge, ...) introduced in #179 .

Also removes benchmarks (few mass matrix UFL files were moved to /demo/) - not used for longer.

Dolfinx tested here https://github.com/FEniCS/dolfinx/commits/michal/ffcx


francesco-ballarin:
Thanks @michalhabera

PR #188, opened 2020-01-10
Added quadrature permutations

mscroggs:
Added quadrature permutations so that FFC can correctly permute quadrature points to make them agree on a facet when viewed from either side.

This is necessary so that meshes no longer need to be ordered by dolfin. By not ordering the meshes, we lose the assumption that the orientations of facets from either side agree, so permutations must be used to correct this.

This method also works on quads, where ordering the mesh does not always work and is infeasible in parallel.

For facet integrals, an array of two numbers will be passed into the `quadrature_permutation` argument of `tabulate_tensor`. The first number is for the first cell adjacent to the facet, the second is for the second cell. These numbers tell FFC which permutations need to be applied to get the facet to the orientation where:
- the lowest numbered vertex is at the origin
- (for 2D facets) the next vertex anticlockwise from the origin has a lower number than the next vertex clockwise from the origin.

For 2D cells, the permutation number will be 0 if no permutation is needed, 1 is a reflection is needed. For 3D cells, N%2 gives the number of reflections needed and floor(N/2) gives the number of rotations needed.

In the generated code, the precomputed values of basis functions and precomputations have an additional dimension for the permutation: these values are calculated for each rotation and reflection of the quadrature points, then the relevant one is used for each facet (eg `FE11_C0_F_Q4[quadrature_permutation[0]][...]`).

This pull request is coupled to https://github.com/FEniCS/dolfinx/pull/696.


ffc/codegeneration/ufc.h
```diff
@@ -413,7 +413,8 @@ extern "C"
       const ufc_scalar_t* c,
       const double* restrict coordinate_dofs,
       const int* entity_local_index,
-      const int* cell_orientation);
+      const int* cell_orientation,
+      const int* quadrature_permutation);
```
michalhabera: Please add a thorough documentation into the docstring above. The docstring above is important for anyone willing to comply with FEniCS interface.

mscroggs: Done. I think cell_orientation is no longer used anywhere, but I need to check this and if so will remove it in a separate PR.

mscroggs:
Closing this as the changes here are a subset of those in https://github.com/FEniCS/ffcx/pull/190

PR #190, opened 2020-01-24
Quadrature point permuting for facets and vector dof directions

mscroggs:
This PR containts all https://github.com/FEniCS/ffcx/pull/188 plus stuff to make non-scalar dofs work on non ordered meshes.

>Added quadrature permutations so that FFC can correctly permute quadrature points to make them agree on a facet when viewed from either side.
>
>This is necessary so that meshes no longer need to be ordered by dolfin. By not ordering the meshes, we lose the assumption that the orientations of facets from either side agree, so permutations must be used to correct this.
>
>This method also works on quads, where ordering the mesh does not always work and is infeasible in parallel.
>
>For facet integrals, an array of two numbers will be passed into the `quadrature_permutation` argument of `tabulate_tensor`. The first number is for the first cell adjacent to the facet, the second is for the second cell. These numbers tell FFC which permutations need to be applied to get the facet to the orientation where:
>- the lowest numbered vertex is at the origin
>- (for 2D facets) the next vertex anticlockwise from the origin has a lower number than the next vertex clockwise from the origin.
>
>For 2D cells, the permutation number will be 0 if no permutation is needed, 1 is a reflection is needed. For 3D cells, N%2 gives the number of reflections needed and floor(N/2) gives the number of rotations needed.
>
>In the generated code, the precomputed values of basis functions and precomputations have an additional dimension for the permutation: these values are calculated for each rotation and reflection of the quadrature points, then the relevant one is used for each facet (eg `FE11_C0_F_Q4[quadrature_permutation[0]][...]`).
> (from https://github.com/FEniCS/ffcx/pull/188)

Additionally, FFC accepts bools when transforming basis functions from the reference element. In the generated code, these are used to multiply dofs that represent vector functions by -1 to ensure that their directions are correct.

If a space contains no vector does, the bool input is ignored.

A version of https://github.com/FEniCS/ffcx/pull/182 is included in this PR as the types of each dof are required.

This pull request is coupled with https://github.com/FEniCS/dolfinx/pull/702.


ffc/codegeneration/finite_element.py
```diff
@@ -307,34 +307,93 @@ def transform_reference_basis_derivatives(L, ir, parameters):
...
-                    L.ForRanges((r, 0, num_derivatives_g),
-                                index_type=index_type,
-                                body=[
-                                    L.AssignAdd(values[ip, idof, r, physical_offset + i],
-                                                transform[r, s] * mapped_value)])
-                ])
-        ]
 
+        # If at least one vector dof needs reflecting
+        if reflect:
```
michalhabera: Can you reduce the code duplication here, below? If reflect=True then prepare the conditional statement, otherwise just multiply with literal 1.0 for example.

ffcx/codegeneration/ufc.h
```diff
@@ -406,14 +408,24 @@ extern "C"
   ///         orientation of the whole mesh.
   ///         0 means "up"
   ///         1 means "down" and scales det(J) with -1.0
-  ///         Applies to the case of k-dimensional surface in n-dimensional space, where k < n.
+  ///         Applies to the case of k-dimensional surface in n-dimensional
+  ///         space, where k < n.
+  /// @param[in] quadrature_permutation
```
garth-wells: Add comment on the size of the array.

PR #201, opened 2020-02-14
Quadrature permutations for DG and some negating of vector-valued spaces 

mscroggs:
- Added quadrature permutations to permute the quadrature points to make DG work.
 - Accept facet rotation and reflection data from dolfinx so these can be used to calculate quadrature permutations and vector direction swaps
 - Moved dof permutations into FFC, as these will likely eventually be calculated using data directly from FIAT.
 - Separated out the reverse engineering based on dof types into `dof_permutations.py`
 - Read `UFC_HEADER_DECL` from `ufc.h` instead of duplicating it.
 - High order vector-valued spaces are not yet supported on unordered meshes, warnings to say this are added.
 - `entity_block_size` is removed, as this is not constant for higher order vector spaces. Instead, this functionality is now included in the calculate of the base dof permutations.

This pull request resolves FEniCS/dolfinx#742, and partially covers FEniCS/dolfinx#736 and  FEniCS/dolfinx#737

This is an updated version of #190, and is twinned with FEniCS/dolfinx#770.


ffcx/codegeneration/definitions.py
```diff
@@ -11,6 +11,8 @@
 
 from ffcx.fiatinterface import create_element
 
+# from ffcx.codegeneration.utils import get_vector_reflection
```
michalhabera: Leftover commented code

ffcx/codegeneration/definitions.py
```diff
@@ -189,12 +189,8 @@ def jacobian(self, e, mt, tabledata, num_points, access):
         return self._define_coordinate_dofs_lincomb(e, mt, tabledata, num_points, access)
 
     def cell_orientation(self, e, mt, tabledata, num_points, access):
```
michalhabera: Is this even needed? I mean this is now too trivial, is it possible to eliminate any reference to this variable at first?

ffcx/codegeneration/ufc.h
```diff
@@ -103,19 +123,28 @@ extern "C"
...
         const double* restrict X);
 
+    /// Map the reference values on to the cell
+    /// @param[in] edge_reflections An array of bools to say whether or not each
+    /// each edge has been reflected. This is used to ensure that vector dofs
+    /// are correctly oriented.
+    /// @param[in] face_reflections An array of bools to say whether or not each
+    /// each face has been reflected. This is used to ensure that vector dofs
+    /// are correctly oriented.
+    /// TODO: Use std::vector<int32_t> to store 1/0 marker for each edge/face
```
michalhabera: What do you mean with std::vector? FFCX and UFC is pure C.

ffcx/codegeneration/ufc.h
```diff
@@ -382,52 +396,66 @@ extern "C"
...
+  ///         this will have size 1.
+  /// @param[in] edge_reflections An array of bools to say whether or not each
+  ///         each edge has been reflected. This is used to ensure that vector
+  ///         dofs are correctly oriented. This array will have one entry for
+  ///         each edge of the cell.
+  /// @param[in] face_reflections An array of bools to say whether or not each
+  ///         each face has been reflected. This is used to ensure that vector
+  ///         dofs are correctly oriented. This array will have one entry for
+  ///         each face of the cell.
+  /// TODO: Use std::vector<int32_t> to store 1/0 marker for each edge/face
```
michalhabera: Same here for std::vector.

ffcx/codegeneration/ufc.h
```diff
@@ -382,52 +396,66 @@ extern "C"
...
+  typedef void(ufc_tabulate_tensor)(
+      ufc_scalar_t* restrict A, const ufc_scalar_t* w, const ufc_scalar_t* c,
+      const double* restrict coordinate_dofs, const int* entity_local_index,
+      const uint8_t* restrict quadrature_permutation,
+      const bool* edge_reflections, const bool* face_reflections,
+      const uint8_t* face_rotations);
```
michalhabera: You didn't document `face_rotations`. Also, it feels a bit weird now. If I understand correctly, in general, for vector dofs you need to know `face_rotations` for a cell integrals too, right? So for a interior facet integrals, isnt now information about rotations of codim 1 entities, facets, duplicated?

ffcx/codegeneration/definitions.py
mscroggs: I've removed cell_orientation

ffcx/codegeneration/ufc.h
mscroggs: `quadrature_permutation` has information about the rotation from two cells either side, while the other inputs are information about the different entities on one cell. But I agree that restructuring what these inputs are at some point is a good idea.

ffcx/codegeneration/ufc.h
mscroggs: These comments were meant to be in dolfin (eg here: https://github.com/FEniCS/dolfinx/blob/ca87bfcf17745f301e4b664871eddae1e02f8619/cpp/dolfinx/mesh/Topology.h#L122), and were included in FFC also by mistake. I've removed them

ffcx/codegeneration/utils.py
```diff
@@ -65,3 +65,89 @@ def generate_return_literal_switch(L,
...
+                # Mark this space as needing reflections
+                _vnames_to_reflect[vname] = True
+
+    # If no dofs need reflecting, don't write any array
+    if not _vnames_to_reflect[vname]:
+        return []
+
+    # If at least one vector dof needs reflecting
+    return [L.ArrayDecl(
+        "const bool", L.Symbol(vname), (len(reflect_dofs), ), values=reflect_dofs)]
```
michalhabera: You can declare the array as `static const bool` instead. FE tables are also static, so the thread-safety is anyway not guaranteed. Similar for dofmap array below.

ffcx/ir/dof_permutations.py
```diff
@@ -0,0 +1,438 @@
...
+    """Permute the dofs on a tetrahedron."""
+    return tetrahedron_rotations(dofs, blocksize) + [tetrahedron_reflection(dofs, blocksize, reverse_blocks)]
+
+
+def permute_hexahedron(dofs, blocksize, reverse_blocks=False):
+    """Permute the dofs on a hexahedron."""
+    return hexahedron_rotations(dofs, blocksize) + [hexahedron_reflection(dofs, blocksize, reverse_blocks)]
+
+
+def empty_permutations(num_perms, num_dofs):
```
michalhabera: I think the usual/better term would be `identity_permutations`.

PR #213, opened 2020-02-24
Use int

michalhabera:


garth-wells:
Could you double-check that UFC no longer has a function that computes a 'full' dofmap based on mesh entity indices?

Hopefully that is now all gone, but if it's still there we need to be careful with the int type if there is a function that takes global mesh entity indices.

michalhabera:
UFC.h is using `int` only. And I guess this would scream already if compiling against ufc and passing `int64_t`.
Anyway I checked and there is no use of `int` in UFC that would require a global index.

PR #221, opened 2020-02-28
Add CFFI JIT warning

michalhabera:


garth-wells:
Some logging would be helpful - will come back as something other than a 'warning'.

PR #226, opened 2020-03-12
Remove block modes

michalhabera:
Removes block modes and scatters "blocks" directly into final tensor A[] within quadrature loop.
Simplifies the code generation and generated code significantly.

The reasons behind this change is following:
1. This PR generates only one code path. It was very hard in past to implement new features and make sure every block mode would work - and even see under what assumptions would these block modes activate etc.
2. Preintegration and premultiplication did a lot of work for compiler. The issue is, that it mostly applied preintegration also to any case when possible *without any more heuristics and logic* whether the actual precomputation are beneficial or not. This might, and it does, depend on the actual cpu architecture and compilers are already doing good job with these decisions.

For example, a very simple Poisson P1 triangle kernel differs in the following line:

current ffcx master, preintegrated block mode:
```C
    // UFLACS block mode: preintegrated
    A[0] += 0.5 * sp[19] + 0.5 * sp[18] + 0.5 * sp[18] + 0.5 * sp[17];
    A[1] += -0.5 * sp[19] + -0.5 * sp[18];
    A[2] += -0.5 * sp[18] + -0.5 * sp[17];
    A[3] += -0.5 * sp[19] + -0.5 * sp[18];
    A[4] += 0.5 * sp[19];
    A[5] += 0.5 * sp[18];
    A[6] += -0.5 * sp[18] + -0.5 * sp[17];
    A[7] += 0.5 * sp[18];
    A[8] += 0.5 * sp[17];
```

while this PR produces:
```C
    // Only 1 quadrature point, no loop
    {
        const ufc_scalar_t fw0 = sp[19] * weights1[0];
        for (int i = 0; i < 2; ++i)
            for (int j = 0; j < 2; ++j)
                A[3 * i + j] += fw0 * FE3_C0_D10_Q1[0][0][0][i] * FE3_C0_D10_Q1[0][0][0][j];
        const ufc_scalar_t fw1 = sp[18] * weights1[0];
        for (int i = 0; i < 2; ++i)
            for (int j = 0; j < 3; ++j)
                A[3 * i + j] += fw1 * FE3_C0_D10_Q1[0][0][0][i] * FE3_C0_D01_Q1[0][0][0][j];
        for (int i = 0; i < 3; ++i)
            for (int j = 0; j < 2; ++j)
                A[3 * i + j] += fw1 * FE3_C0_D01_Q1[0][0][0][i] * FE3_C0_D10_Q1[0][0][0][j];
        const ufc_scalar_t fw2 = sp[17] * weights1[0];
        for (int i = 0; i < 3; ++i)
            for (int j = 0; j < 3; ++j)
                A[3 * i + j] += fw2 * FE3_C0_D01_Q1[0][0][0][i] * FE3_C0_D01_Q1[0][0][0][j];
    }
```
What any good compiler could under `-ffast-math` already do, is to decide whether loops should be unrolled (they are in this case) and inline static tables into the computations - which results basically in the preintegrated mode.
Current clang 9.0.0 compiler with `-Ofast -march=haswell` produces for both code version ~ 55 instructions without loops (https://godbolt.org/z/LVGF8z).


ffcx/codegeneration/integrals_generator.py
```diff
@@ -257,8 +238,6 @@ def generate_element_tables(self):
 
         tables = self.ir.unique_tables
         table_types = self.ir.unique_table_types
-        inline_tables = self.ir.integral_type == "cell" and not (self.ir.needs_rotations or self.ir.needs_reflections)
```
mscroggs: This (and the same code on line 1095) is the only place that I used `needs_rotations` and `needs_reflections`, so these can now be removed from the ir (probably in a future PR not here?)

ffcx/ir/uflacs/elementtables.py
```diff
@@ -571,19 +564,6 @@ def optimize_element_tables(tables,
...
-        uname = unique_names[ui]
-
-        # Track table origins for runtime recomputation in custom integrals:
-        name = "name of 'smallest' element we can use to compute this table"
-        dofrange = "FIXME"  # table_ranges[name]
-        dofmap = "FIXME"  # table_dofmaps[name]
-
-        # FIXME: Make sure the "smallest" element is chosen
-        (element, avg, derivative_counts, fc) = table_origins[name]
-        unique_table_origins[uname] = table_origin_t(element, avg, derivative_counts, fc, dofrange,
```
mscroggs: I think table origins can also be removed from the ir as it was only used to work out how to add permutations to preintegrated tables. (Again this is probably to be done in a future PR)

PR #232, opened 2020-04-07
Make RTCF spaces work on quad meshes

mscroggs:
This makes order 1 RTCF div-conforming spaces on quads work. RTCF1 is shown here on the right:
RTc1

~Currently this only works on quad meshes that are ordered so that the orientations of edges on adjacent quads agree.~ This now works on quad meshes with non-matching edge directions.

This PR coupled with FEniCS/dolfinx#946.


mscroggs:
Closing for now until bugs are fixed

Issue #234, opened 2020-04-10
JIT is leaking memory


garth-wells:
The JIT compiler call a factory function from the compiled UFC code, see https://github.com/FEniCS/ffcx/blob/e492492b01e1e2788f18c40dbabd329117ab3727/ffcx/codegeneration/jit.py#L324. However, Python and cffi cannot know that this is factory function, hence cannot know that it needs to be freed [1]. 

The JIT need to also return a function that wraps `free(void*)` and make clear that the caller needs to manage the lifetime. Alternatively, the object created by the factory function could be wrapped in a Python class with a destructor that frees the memory.

[1] https://stackoverflow.com/questions/47080411/do-i-need-to-free-memory-returned-from-a-c-function-called-via-cffi

michalhabera:
Solved via #235

PR #235, opened 2020-04-12
Fix JIT memory leak

michalhabera:
Fix for #234, follows suggestion on https://stackoverflow.com/questions/47080411/do-i-need-to-free-memory-returned-from-a-c-function-called-via-cffi

This exposes standard C `free()` and wraps UFC create_foo() factory result using `ffi.gc()`.


garth-wells:
See related https://github.com/FEniCS/dolfinx/issues/966 which should ideally be fixed at the same time.

PR #237, opened 2020-04-14
Defer importing FIAT

michalhabera:
`FIAT` is very expensive to import, mainly because it imports `sympy`. Run `python3 -X importtime -c "import ffcx;"` to see.

In a typical use-case from dolfinx, where every code is cached and precompiled there is absolutely no need to import FIAT.

This gives speed-up more than 30% for hot runtime of small size problems.
Current master, incl. importing FIAT:

```
time python3 small_problem.py

real	0m3.134s
user	0m1.022s
sys	0m0.528s
```

With this PR:

```
time python3 small_problem.py

real	0m1.791s
user	0m0.579s
sys	0m0.456s
```

### DIFF
--- a/ffcx/__init__.py
+++ b/ffcx/__init__.py
@@ -13,24 +13,11 @@ FFC compiles finite element variational forms into C code.
 import logging
 import pkg_resources
 
-from FIAT import supported_elements
-
 __version__ = pkg_resources.get_distribution("fenics-ffcx").version
-
 
 logging.basicConfig()
 logger = logging.getLogger("ffcx")
 logging.captureWarnings(capture=True)
 
-# Import main function, entry point to script
-from ffcx.main import main  # noqa: F401
-
 # Import default parameters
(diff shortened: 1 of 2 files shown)


ffcx/__init__.py
```diff
@@ -13,24 +13,11 @@
...
 
-# Import main function, entry point to script
-from ffcx.main import main  # noqa: F401
-
 # Import default parameters
 from ffcx.parameters import default_parameters  # noqa: F401
-
-# Duplicate list of supported elements from FIAT and emove elements from
-# list that we don't support or don't trust
-supported_elements = sorted(supported_elements.keys())
```
garth-wells, line 33: Should this be handed elsewhere?

michalhabera: It is handled. It was done twice in a weird way. FIAT has supported elements which clearly says which elements it supports, we should not monkey patch it.

Supported elements by FFCx are stated in `fiatinterface.py` as `supported_families`.

PR #244, opened 2020-05-27
Clean quadrature degrees/schemes

michalhabera:
Major clean-up of quadrature rule/degree/scheme in ffcx.

In past, each integral group had several quadrature rules identified by the number of points in the quadrature group. This didn't allow to have different quadrature schemes applied within the same integral group.
This PR identifies integral groups for integrands with a `dict` which maps `QuadratureRule` class into the integrand expression. `QuadratureRule` now serves as an object with a unique hash and a deterministic ID. This object could in future be extended to allow runtime quadrature, it can possibly serve as an interface with dolfinx to specify the runtime rules.

After this PR only very little is needed to be able to hardcode custom compile-time quadrature rules, will come very soon in a separate PR.

This PR also benefits from better quadrature degree estimation in https://github.com/FEniCS/ufl/pull/27


michalhabera:
Any thoughts @mscroggs @chrisrichardson @garth-wells ? I'd like to proceed this/next week, custom compile-time quadrature rules should a matter of a simple PR on top of this.

garth-wells:
@michalhabera Does this PR change any current behaviour, or does it just open new possibilities?

michalhabera:
@garth-wells It is mostly a clean-up and refactorization. It changes naming within the generated code. Because in past any variable associated with a quadrature rule used just number of quadrature points to define its unique name.
E.g. a FE table used in 3-point quadrature rule was named `FE3_C0_D01_Q3`, where `Q3` means 3-point quadrature.
New code would name it `FE3_C0_D10_Q48e` where `48e` is just a unique (deterministic) ID given to `QuadratureRule` object computed from the hash of array of points and weights.
And this naming change now holds for almost any variable, instead of `sv3` we have now `sv_48e`, ...
This PR also stores only one factorization graph. In the past, there was "quadrature independent" and "quadrature dependent graph" stored. But this is not needed, because information, whether an expression is dependent on quadrature rule used, is already stored in the graph.

PR #247, opened 2020-05-29
Save the cffi build command to the status-ready file

chrisrichardson:
Small change to save the compilation command to file. 
There is already a (zero-size) lock file, which is created on compilation, so that seems a logical place to put it. This can be useful when debugging generated code:

e.g. 
- go to the cache directory
- edit libffcx_cmaps_14b073cae1764e0bcfd32bb297be3687834d72a9.c
- `tail -2 libffcx_cmaps_14b073cae1764e0bcfd32bb297be3687834d72a9.c.cached  | bash` will recompile.

### DIFF
--- a/ffcx/codegeneration/jit.py
+++ b/ffcx/codegeneration/jit.py
@@ -4,7 +4,9 @@
 #
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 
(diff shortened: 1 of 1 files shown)


chrisrichardson:
Happy to change name, but not sure what to.

Issue #236, opened 2020-04-13
Add ufc_free function to free memory


garth-wells:
UFC has a factory-style where `create_foo` functions dynamically allocate memory. This needs to be partnered with a `ufc_free` function that frees the memory.

A present we're calling `free` to destroy the UFC objects, but this assumes that they have been allocated with `malloc`, which is an implementation detail.

garth-wells:
No required with new static design.

PR #249, opened 2020-05-29
Init and destroy functions for ufc objects

jamtrott:
This pull requests adds a number of functions to the UFC api in `ufc.h`.

Previously, objects, such as `ufc_dofmap`, `ufc_form`, and so on, could be created by the user by calling the corresponding `create()` function. Such a function returns a pointer to the corresponding object. Internally, the `create()` function allocates storage for the object using `malloc()` and then sets the appropriate fields of the corresponding struct. The user is required to free the allocated memory by calling `free()` whenever they are done.

In these patches, I have added additional functions to each of the objects in the UFC API, so that they can be created without having UFC allocate storage. Instead, the user provides a pointer to the object and an `init()` function is used to initialise the members of the struct. In addition, I have added corresponding `destroy()` functions, which are intended to be used to free any resources that are associated with the object. As of now, `destroy()` does nothing, since there are no resources associated with any objects. However, I have in mind a future use case, where such functions are needed.

This pull request addresses Issue #236.

### DIFF
--- a/ffcx/codegeneration/coordinate_mapping_template.py
+++ b/ffcx/codegeneration/coordinate_mapping_template.py
@@ -4,12 +4,16 @@
 # The FEniCS Project (http://www.fenicsproject.org/) 2018
 
 declaration = """
+int init_{factory_name}(ufc_coordinate_mapping* cmap);
+void destroy_{factory_name}(ufc_coordinate_mapping* cmap);
 ufc_coordinate_mapping* create_{factory_name}(void);
 
-// Helper used to create coordinate map using name given to the
-// UFL file.
-// This helper is called in user c++ code.
+// Helper functions used to create coordinate maps
+// based on the name of the UFL file.
+// These helpers are called in user c++ code.
 //
+int init_coordinate_map_{prefix}(ufc_coordinate_mapping* cmap);
+void destroy_coordinate_map_{prefix}(ufc_coordinate_mapping* cmap);
 ufc_coordinate_mapping* create_coordinate_map_{prefix}(void);
 """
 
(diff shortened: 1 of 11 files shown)


garth-wells:
Could you expand on why the `init` functions might be useful?

jamtrott:
At the moment, the `init()` functions are exactly like the corresponding `create()` functions, except that they do not call `malloc()`. Instead, the user passes in a pointer to the object. So, to begin with, it is mostly useful if you would like to avoid  unnecessary calls to `malloc()` and `free()` in cases where the UFC objects can have automatic storage duration.

In the long term, I'm planning to use the `init()` and `destroy()` functions to manage some resources that I would like to associate with the UFC objects. More specifically, I'd like to offload certain functions to a GPU, and this implies managing resources that are needed to compile device-side (GPU) code and launch them as kernels.

Perhaps it could be useful in some other cases too. For instance, memory could be allocated in `init()` as a temporary workspace that is used in, for example, `tabulate_tensor()`. However, I don't have any concrete use cases like that. In any case, it allows the FFC-X implementation some more flexibility, while encapsulating implementation details, hiding them from the user.

garth-wells:
I would suggest removing the `init` functions - we've been consciously removing anything that is not actively used because every addition comes with a maintenance and testing burden.

We can look again in the future if there is a concrete use case. I think it's a bit early to speculate on what might be required for GPUs, etc. We'd need to see examples to make the case.

jamtrott:
Thanks for your comments, Garth. Sure, I can appreciate the need to remove unused code.

I think my earlier explanation only served to confuse. Let’s forget about GPUs, and, if you don’t mind, I will have another try at explaining.

I want to change the signature of the `create()` functions to take a pointer to the relevant object as an argument, rather than returning a pointer to the object. That way, the code generated by FFCX no longer has to call `malloc()` to allocate storage for the object. Instead, it is the responsibility of the caller (i.e., Dolfin-X) to provide the storage. Note that this resolves Issue #236.

However, the `create()` functions are part of the user-facing API of FFCX. Now, I’m not sure whether there is an accepted process for changing the API or if there are any requirements related to compatibility. In any case, I thought it would be a good idea to not change the `create()` functions directly. Instead, I added the `init()` functions, which have the new signature. If the users of the FFCX API were all changed to use the proposed `init()` functions, then it would be safe to deprecate and remove the old `create()` functions.

In the end, this does perhaps seem like a lot of work for little gain. If the proposed API is not considered an improvement, or if there is a better way to resolve Issue #236, then please feel free to close this pull request.

michalhabera:
@jamtrott What would be an advantage of dolfinx passing a preallocated memory? Just keep in mind that the UFC objects are never heavily used in performance-critical code, neither there is ton of them.
This would be a nice change for objects for which we'd like to control memory alignment. But as far is I can tell, this is not the case for UFC objects. These are allocated once and basically copied into dolfinx data structures.

jamtrott:
@michalhabera, you're right. This is not going to impact performance. The only advantage would be to avoid calls to `malloc()` and `free()`. Instead you could just use variables with automatic storage duration.

At the moment, dolfinx does something like the following:
```
ufc_form form = ...;
ufc_integral * integral = form.create_cell_integral(id); // ffcx calls malloc()
// ... Copy data from ufc_integral ...
free(integral); // User is required to call free()
```

Now, it would look something like this instead:
```
ufc_form form = ...;
ufc_integral integral;
form.init_cell_integral(&integral, id);
// ... Copy data from ufc_integral ...
```

michalhabera:
@jamtrott I see, however it seems like a lot of code just for this little improvement. Currently, user is calling `create_foo` and we should just document if it is not done, that it `malloc`s a memory. The real issue #236 of exposing ufc_free function could be resolved simpler, with a few lines of code.

Do you want to try a PR for that?

Note: As you can see in other PRs, work is mainly in the direction of code clean-up and removal. We have been trying to cut loads of unused and complicated code paths.

jamtrott:
Sure. Thank you both for your comments. I will open a new pull request with a patch including the documentation you suggest. This essentially amounts to keeping the first patch in the current pull request and dropping the rest.

PR #250, opened 2020-05-30
Get compute_reference_coordinates working for quad/hex

chrisrichardson:
* fix critical bug in dphi ordering
* adjust tolerance for Newton method
* add tests


ffcx/codegeneration/coordinate_mapping.py
```diff
@@ -204,7 +204,7 @@ def compute_physical_coordinates(L, ir):
 
 def compute_reference_geometry(L, ir):
     degree = ir.coordinate_element_degree
-    if degree == 1:
+    if degree == 1 and ir.cell_shape in ("interval", "triangle", "tetrahedron"):
```
michalhabera, line 207: We should probably prepare a more standard way of telling if a coordinate map is affine, maybe have `ir.affine = True`?

test/test_jit_cmaps.py
```diff
@@ -19,13 +19,15 @@ def test_cmap_triangle(compile_args):
...
+
+    # Reference coordinates X
+    X = np.array([[0.5, 0.5]], dtype=np.float64)
     X_ptr = module.ffi.cast("double *", module.ffi.from_buffer(X))
+    x = np.zeros_like(X)
+    x_ptr = module.ffi.cast("double *", module.ffi.from_buffer(x))
     coords = np.array([0.0, 0.0, 2.0, 0.0, 0.0, 4.0], dtype=np.float64)
     coords_ptr = module.ffi.cast("double *", module.ffi.from_buffer(coords))
-    compiled_cmap[0].compute_reference_coordinates(X_ptr, X.shape[0], x_ptr, coords_ptr)
+    compiled_cmap[0].compute_reference_coordinates(x_ptr, x.shape[0], X_ptr, coords_ptr)
```
michalhabera, line 30: Why like this? `ufc.h` says 
```c
    void (*compute_reference_coordinates)(
        double* restrict X, int num_points, const double* restrict x,
        const double* restrict coordinate_dofs);
```
where capitalized always means reference.

ffcx/codegeneration/coordinate_mapping.py
```diff
@@ -447,13 +445,10 @@ def _compute_reference_coordinates_newton(L, ir, output_all=False):
...
-    #       e.g. is epsilon=1e-6 and iterations=degree sufficient?
-    max_iter = L.LiteralInt(degree)
-    epsilon = L.LiteralFloat(1e-6)
-    # TODO: Could also easily make criteria input if desired
-    # max_iter = L.Symbol("iterations")
-    # epsilon = L.Symbol("epsilon")
+    # FIXME: return an error code (e.g. -1) on non-convergence
+    # Max ten iterations seems plenty, epsilon^2=1e-12
+    max_iter = L.LiteralInt(10)
+    epsilon2 = L.LiteralFloat(1e-12)
```
michalhabera, line 451: Ok for a quick-fix, but in future we should have a way of controlling this. Maybe passing these as arguments to `compute_reference_coordinates`?

garth-wells: We need at least an error code before merging.

test/test_jit_cmaps.py
chrisrichardson: That's what I thought too. I must have made a mistake.

PR #251, opened 2020-05-30
Store if affine in coordinate map IR

michalhabera:


garth-wells:
I think 'is_affine` would be clearer.

Issue #257, opened 2020-06-04
Provide rank (in tensor sense) of an expression in generated code


mrambausek:
Currently, ffcx generates a "value_shape" attribute as a fixed size array for compiled expressions. However, there is no information about the length of this array which makes the interface incomplete. 
Note that from python one can access the value_shape array beyond its actual bounds without any problems (just getting a zero). I suppose cffi does some magic here to treat this properly.
As a second point, for scalars (rank zero) ffcx currently generates 
```
static const int value_shape[0] = { 0 };
```
which gives a compiler warning.

mrambausek:
Did not see that there is num_components which does the job. The second point of the comiler warning then is a different story.

mrambausek:
Sorry for spamming on this

michalhabera:
Can you create a short and simple PR which would just avoid compiler warning for scalar values expressions?

mrambausek:
Yes, working on it. What do you think about the name of num_components? I find it confusing since my first interpretation would some kind of size of the actual data array but not of the shape array.

PR #258, opened 2020-06-04
Change the detection of a zero-valued array: empty is not zero

mrambausek:
By that, the generated code for an empty array will feature an empty initializer list instead of `{ 0 }` and thus avoid a corresponding compiler warning.


garth-wells:
@mrambausek could you expand on the the impact this will have on the initializer list? Empty initializer lists are not valid C.

mrambausek:
@garth-wells thanks for pointing that out. Indeed, my PR leads to empty initializers
```
static const int value_shape[0] = {  };
```
I did not know that this is not allowed in C but assumed that it is valid since gcc did not complain.

michalhabera:
Maybe we can change a declaration of value_shape to be `static const int value_shape[1] = {0};` for scalars? Num_components is just telling that it is a scalar and value_shape should not be accessed.
A difference between shape `()` and `(0,)` is anyway rather dull.

mrambausek:
@michalhabera you are right. I guess, your suggestion is the only standard conforming one. If you and @garth-wells are okay with that, I'll update this PR accordingly -- also looking for other cases that produce empty arrays.

michalhabera:
There is also a similar case in `original_coefficient_position` for form. But there the logic is a bit different - there is a method which returns value at position `i`. I.e. the issue of not having the size of the array in the UFC header is bypassed with a method.

Here, memory goes into static array and UFC structure holds just a pointer. So another way how to fix this would be to follow the "method" way and return always `-1` in case of scalar-valued expression.

garth-wells:
See on `{}` https://stackoverflow.com/questions/17589533/is-an-empty-initializer-list-valid-c-code. GCC allows it - we changed the code to avoid `{}` a while back because other compilers complain.

mrambausek:
@garth-wells thanks for clarifying. Also had a look at the C11 standard which still forbids 
`static const int value_shape[0]` anyway.

mrambausek:
@michalhabera your suggested fix for the array version is non-intrusive, ie. without changes to ufc, but the member function version is appealing for consistency of patterns. What is preferred? If ufc is touched, I'd still favor a better name instead of `num_components`, e.g. `rank`.

garth-wells:
I think we need some clarification on what the problem is. It's not clear to me from the discussion.

mrambausek:
@garth-wells The initial problem was mentioned as a sidenote in Issue #257 (closed). Currently, the code generated for scalar expressions is
```
static const int value_shape[0] = { 0 };
```
which leads to a warning from gcc (and is not standard conforming).

michalhabera:
Yes, you are right the information provided now might be incomplete for a C++ workflow. The reason is that we do not yet have working pipeline for expressions in a typical CLI. But let's not get into this unnecessarily.

I am happy with `static const int value_shape[1] = {0};` for scalars.

mrambausek:
Okay, I went back to the old logic of _is_zero_valued being true for empty arrays and instead make a minimum array length of 1 (per slot)

michalhabera:
I meant to fix this directly in value_shape of expression. I am afraid your fix now could hide an error somewhere else, where we would on empty array rather see a warning.

mrambausek:
@michalhabera I will do so. What do you think about an exception or assertion for that case of zero dimension, which is more explicit than catching this by the compiler.

michalhabera:
@mrambausek Good idea. Can you add an exception for the general case which would lead to invalid C code?

mrambausek:
@michalhabera I added a check in `ArrayDecl.cs_format`. I though about putting it in the constructor but I preferred to have it closer to the place where it becomes apparent as there is no guarantee that sizes will not be modified before the call to `cs_format`. I did not add further checks as there are already a number of assertions in `build_initializer_lists` on the compatibility of "values" with "sizes"

michalhabera:
@mrambausek Just one last: could you move the condition which decides on shape declaration to the body of `generate_value_shape()`? Data type and keywords are there, I would not like to see it spread at too many places.

mrambausek:
@michalhabera I thought about that already before. However, `generate_value_shape()` returns not simply a string but a CStatement. Thus I guess I would need to create a new subclass, e.g., EmptyArrayDecl [Some dummy class with a `cs_format()` method would do as well but is also hackish], since you did not want to have the code in ArrayDecl either. Is this the way to go?

mrambausek:
I could of course also let `generate_value_shape` return a `str` instead of a `CStatement`, but that would break the pattern in `ExpressionGenerator`

PR #259, opened 2020-06-04
Fix issue 252: handle the case of an expression that is a compile-time constant

mrambausek:
Here I propose to generate code for compile-time constants. The use case is to support for example derivatives that lead to expressions without any ufl coefficients and constants left, e.g.,
```
expr = ufl.diff(ufl.diff(f*f, f), f)
```
where `f` is some ufl.Coefficient. Something like this is quite natural to occur for library code where there is no control over the given expression of which the derivative has to be compiled.
I am not sure whether this PR is the most reasonable way to handle such sitatuations -- any suggestions? Is there any other way to evaluate this besides going through ffcx/uflacs?


mrambausek:
The failure is due to coverall upload having a problem....

michalhabera:
The fix is OK, but the test is probably too much boilerplate for such little actually testing. Could you remove the test?

PR #261, opened 2020-06-06
Custom compile time quadrature rules

michalhabera:
Allows a user to pass his own set of points and weights as metadata in integral.

The syntax is following
```python
points = [[0.0, 0.0], [1.0, 0.0], [0.0, 1.0], [0.5, 0.5], [0.0, 0.5], [0.5, 0.0]]
weights = [1 / 12] * 6
u * v * ufl.dx(metadata={"quadrature_rule": "custom", "quadrature_points": points, "quadrature_weights": weights})
```

### DIFF
--- a/ffcx/ir/representation.py
+++ b/ffcx/ir/representation.py
@@ -19,6 +19,7 @@ representation under the key "foo".
 import itertools
 import logging
 from collections import namedtuple
+import warnings
 
 import numpy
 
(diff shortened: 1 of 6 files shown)


ffcx/ir/representation.py
```diff
@@ -422,8 +423,37 @@ def _compute_integral_ir(form_data, form_index, prefix, element_numbers, integra
             scheme = md["quadrature_rule"]
             degree = md["quadrature_degree"]
 
-            (points, weights) = create_quadrature_points_and_weights(integral_type, cell, degree,
-                                                                     scheme)
+            if scheme == "custom":
+                points = md["quadrature_points"]
+                weights = md["quadrature_weights"]
+            elif scheme == "vertex":
+                # FIXME: Could this come from FIAT?
```
miklos1, line 430: Quite certainly yes, and it should.

ffcx/ir/representation.py
```diff
@@ -422,8 +423,37 @@ def _compute_integral_ir(form_data, form_index, prefix, element_numbers, integra
             scheme = md["quadrature_rule"]
             degree = md["quadrature_degree"]
 
-            (points, weights) = create_quadrature_points_and_weights(integral_type, cell, degree,
-                                                                     scheme)
+            if scheme == "custom":
+                points = md["quadrature_points"]
+                weights = md["quadrature_weights"]
```
chrisrichardson, line 428: Should there be some checks here, (1) that the points and weights have been provided (and if not, return a useful message), (2) that they are the same length?

mrambausek:
As I am heavily relying on QuadratureElement, are there any plans/ideas of extending them to custom quadrature rules?

michalhabera:
@mrambausek We need to review "QuadratureElement" codepath again. I can have a look at custom points in that case as well.

PR #262, opened 2020-06-06
ufc.h: Document that memory is allocated with malloc() and should be freed by the caller

jamtrott:
This patch adds some comments to the UFC API, `ufc.h`, to document that memory is allocated with `malloc()` and that the user is responsible for calling `free()`.


michalhabera:
@garth-wells Do you know how to resolve failing Coveralls action? We have few PRs from remote origins which could be merged.

michalhabera:
Thanks @jamtrott !

PR #265, opened 2020-06-07
Add parameters documentation as a string

michalhabera:
A more systematic approach to document parameters. Description is now printed as help for CLI, e.g.
```
  --assume_aligned ASSUME_ALIGNED
                        Assumes alignment (in bytes) of pointers to tabulated tensor, coefficients and constants array. This value must be compatible with alignment of data structures
                        allocated outside FFC. (default=4)
```

### DIFF
--- a/ffcx/analysis.py
+++ b/ffcx/analysis.py
@@ -19,7 +19,7 @@ from collections import namedtuple
 import numpy
 
 import ufl
-from ffcx.parameters import FFCX_PARAMETERS
+from ffcx.parameters import default_parameters
 
 logger = logging.getLogger("ffcx")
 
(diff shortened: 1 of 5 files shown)


chrisrichardson:
I thought `assume_aligned` was a bool.

garth-wells:
`assume` is the right word - it's aligned or not aligned.

michalhabera:
It just adds calls `void * __builtin_assume_aligned (const void *exp, size_t align, ...)` extension. So I wouldn't change its name.

And it is not bool. It is the assume alignment of a pointed-to memory. @chrisrichardson you put there a default value as `None`. But our parameter system accepts only values which has the same type as default value. For this reason, I have put there default of `4` which is safe on at least 32-bit machines.

michalhabera:
At least we seen now how important is to document these parameters...

chrisrichardson:
I'm just a bit nervous of having it always there, even if the value is "safe". Could we use 0 or -1 to turn it off.

michalhabera:
@chrisrichardson Yes, good idea. I will take `-1` as disabled.

PR #266, opened 2020-06-08
Simplify by removing intermediate FIAT coordinates

chrisrichardson:
* Simplifies some code generation by not creating intermediate FIAT coordinates in code

### DIFF
--- a/ffcx/codegeneration/evaluatebasis.py
+++ b/ffcx/codegeneration/evaluatebasis.py
@@ -253,6 +253,7 @@ def _generate_compute_interval_basisvalues(L, basisvalues, X, ip, embedded_degre
 
     code += [L.Assign(basisvalues[0], 1.0)]
 
+    # FIAT coordinates
     Y0 = 2 * X[ip] - 1
 
     if embedded_degree > 0:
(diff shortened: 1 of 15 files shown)


ffcx/codegeneration/evaluatebasis.py
```diff
@@ -413,14 +410,14 @@ def _idx2d(p, q):
...
-        code += [L.VariableDecl("const double", f2, (1.0 - Y[1]) / 2.0)]
+        code += [L.VariableDecl("const double", f2, (1.0 - Y1) / 2.0)]
         code += [L.VariableDecl("const double", f3, f2 * f2)]
     for r in range(1, embedded_degree):
         rr = _idx2d((r + 1), 0)
         ss = _idx2d(r, 0)
         tt = _idx2d((r - 1), 0)
-        A = (2 * r + 1.0) / (r + 1)
-        B = r / (1.0 + r)
+        A = 1 + r / (r + 1)
```
michalhabera, line 419: It this correct? Shouldn't it be `2 - 1/(r+1)`?

chrisrichardson: I think it is correct, `(2*r +1)/(r+1) = (r + (r+1))/(r+1) = 1 + r/(r+1)`. But maybe I shouldn't touch it. So many of these expressions are rather cryptic.

michalhabera: Ah, ok, both expressions are correct. Yes, many of these formulas are hard to follow without some additional material...

PR #272, opened 2020-07-13
Remove recursion

michalhabera:
A recursion when replacing `QuadratureWeight` is a major cause for bad performance on complicated forms/expression.

This PR replaces recursion with more efficient tree traversal in UFL.


garth-wells:
Should we test DOLFIN against this change?

michalhabera:
This should have no real effect, but I did test to be sure, https://github.com/FEniCS/dolfinx/compare/michal/ffcx?expand=1

PR #273, opened 2020-07-17
XYZXYZ ordering

mscroggs:
This PR switches VectorElements from using XXYYZZ ordering to XYZXYZ ordering.

This PR is coupled with FEniCS/dolfinx#1088


ffcx/codegeneration/ufc.h
```diff
@@ -180,6 +183,9 @@ extern "C"
     /// for a cell (not including global support dofs)
     int num_element_support_dofs;
 
+    /// Return the block size for a VectorElement
```
garth-wells, line 186: Add explanation for tensor elements.

ffcx/fiatinterface.py
```diff
@@ -89,13 +90,66 @@ def _create_mixed_finiteelement(element: ufl.MixedElement) -> FIAT.MixedElement:
...
+            if isinstance(e, ufl.MixedElement) \
+                    and not isinstance(e, ufl.VectorElement) \
+                    and not isinstance(e, ufl.TensorElement):
                 rextract(e.sub_elements())
             else:
                 elements.append(e)
 
     rextract(element.sub_elements())
     return FIAT.MixedElement(map(_create_element, elements))
+    # return FIAT.MixedElement(map(_create_element, element.sub_elements()))
```
garth-wells, line 102: Remove comment.

ffcx/codegeneration/ufc.h
```diff
@@ -180,6 +184,9 @@ extern "C"
...
+    int submap_block_size;
```
michalhabera, line 188: What does this mean? Why is it needed? Why is block_size in element, but sub_dofmap_block_size in dofmap?

mscroggs: It's needed to handle MixedElements that contain VectorElements. The name is bad: it's used in dolfin to get the blocksize of a subdofmap, but the value itself has nothing to do with sub dofmaps. I'll rename it

ffcx/codegeneration/expressions.py
```diff
@@ -227,34 +227,67 @@ def generate_block_parts(self, blockmap, blockdata):
...
+                if b - a != bm[1] - bm[0]:
+                    unzip = True
+                    break
+            else:
+                continue
+            break
+
+        if unzip:
+            # If DOFs in dofrange are not equally spaced
+            from itertools import product
```
michalhabera, line 243: Import at the top of the file. Deferred imports are good only if there is a very strong reason for.

ffcx/codegeneration/expressions.py
```diff
@@ -227,34 +227,67 @@ def generate_block_parts(self, blockmap, blockdata):
...
+
+            # TODO: handle non-contiguous dof ranges
+
+            A_indices = []
+            for bm, index in zip(blockmap, arg_indices):
+                # TODO: switch order here? (optionally)
+                # TODO: save block size in ir rather than reverse engineering here
+                offset = bm[0]
+                if len(bm) == 1:
+                    ebs = 1
```
michalhabera, line 271: What is `ebs`? Why are you reverse engineering block size here, it is stored in IR.

ffcx/codegeneration/expressions.py
```diff
@@ -227,34 +227,67 @@ def generate_block_parts(self, blockmap, blockdata):
...
+        for i, bm in enumerate(blockmap):
+            for a, b in zip(bm[1:-1], bm[2:]):
+                if b - a != bm[1] - bm[0]:
+                    unzip = True
+                    break
+            else:
+                continue
+            break
+
+        if unzip:
```
michalhabera, line 241: Can you explain what `unzip` means?

ffcx/codegeneration/integrals.py
```diff
@@ -829,17 +825,52 @@ def generate_block_parts(self, quadrature_rule, blockmap, blockdata):
...
+        else:
+            # Fetch code to access modified arguments
+            arg_factors = self.get_arg_factors(blockdata, block_rank, quadrature_rule, iq, B_indices)
+
+            B_rhs = L.float_product([fw] + arg_factors)
+            A_indices = []
+
+            for bm, index in zip(blockmap, arg_indices):
+                # TODO: switch order here? (optionally)
+                # TODO: save block size in ir rather than reverse engineering here
```
michalhabera, line 862: Same comments apply here.

ffcx/ir/elementtables.py
```diff
@@ -207,6 +205,19 @@ def get_ffcx_table_values(points, cell, integral_type, ufl_element, avg, entityt
 
         component_element = fiat_element.elements()[component_element_index]
 
+        # Get the block size to switch XXYYZZ ordering to XYZXYZ
+        if isinstance(ufl_element, ufl.VectorElement) or isinstance(ufl_element, ufl.TensorElement):
+            block_size = len(fiat_element.elements())
```
michalhabera, line 210: FIAT element has num_sub_element, you are calling this somewhere above, would be better to be consistent.

ffcx/ir/dof_permutations.py
```diff
@@ -23,6 +24,14 @@ def base_permutations(ufl_element):
 
     # If the element has sub elements, combine their permutations
     perms = None
+
+    if isinstance(ufl_element, ufl.VectorElement) or isinstance(ufl_element, ufl.TensorElement):
+        block_size = len(ufl_element.sub_elements())
```
michalhabera, line 29: UFL element has num_sub_elements().

ffcx/codegeneration/expressions.py
mscroggs: block size isn't stored in the integral/expression ir, so I'm not sure you can access it here. Also, this allows for elements that have gaps in their dof lists for reasons other than blocksize (which there is probably no current use case, but doing it this way allows for this in future)

PR #279, opened 2020-09-21
Include points in Expression signature computation

jhale:
This PR fixes a subtle bug that results in the same compiled object being called even when the requested points on the reference cell had been changed. This led to incorrect output.


michalhabera (approved):
Thanks!

PR #280, opened 2020-10-06
Strip fewer zeros

mscroggs:
Zeros in a table between non-zero entries should only be stripped when they are due to being components of a blocked element.

Currently, all zeros are stripped, causing for example N1curl's derivatives to all be stripped to all have the same table, then the wrong permutations ending up for some components: this causes FEniCS/dolfinx#1148. (This is a bugfix of a change in #273.)

I've added a bad test to a branch of dolfinx to confirm that this fix is correct:

https://github.com/FEniCS/dolfinx/blob/88be6d9ccedf1586ff75f0ea5aef74f83462abe2/python/test/unit/fem/test_element_integrals.py#L411-L426

I'm currently working on a better test.

### DIFF
--- a/ffcx/ir/elementtables.py
+++ b/ffcx/ir/elementtables.py
@@ -66,7 +66,7 @@ def clamp_table_small_numbers(table,
     return table
 
 
-def strip_table_zeros(table, rtol=default_rtol, atol=default_atol):
+def strip_table_zeros(table, block_size, rtol=default_rtol, atol=default_atol):
     """Strip zero columns from table. Returns column range (begin, end) and the new compact table."""
     # Get shape of table and number of columns, defined as the last axis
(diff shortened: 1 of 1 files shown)


garth-wells (approved):

garth-wells:
Is there a simple way this could be tested in the FFCX tests?

mscroggs:
Yes, I think: we can check that the curls computed for two differently ordered cells are consistent. Currently adding a test along these lines to dolfin, then I'll try to put something similar in ffc

garth-wells:
Will you add a test in this PR or in a separate PR?

mscroggs:
maybe separate, as it's taking a while to get a test working, and it would be helpful for electromagnetics stuff to get this bug fixed?

PR #285, opened 2020-10-27
Parse env parameters

michalhabera:
Adds the possibility to force any parameter with env variable. Especially useful for `export FFCX_ASSUME_ALIGNED=...`. Env forced parameters have the highest priority.

In addition, `verbosity` is added as a new FFCx parameter, which means it could be now easily controlled either via API or the above-mentioned env vars. One can do `export FFCX_VERBOSITY=10`.
It solves the issue of controlling FFCx verbosity from dolfinx, `dolfinx.Form(form, ffcx_parameters={"verbosity": 10})`.


jhale:
Thanks for doing this, I was going to do it so you saved me a job.

With this I can setup the end-user Docker images with 32-bit alignment.

jhale:
Superseded by PR #287.

PR #287, opened 2020-10-30
Configuration files for FFCX

jhale:
This PR greatly simplifies parameter handling in FFCX. It supersedes PR #285.

* passed parameters (command line, API) *always* take precedent over parameters set elsewhere.
* `quadrature_rule`, `quadrature_degree` and `precision` are no longer FFCX compiler parameters. they are exclusively set by the integral metadata (or otherwise set to internal defaults/calculated).
* setting log level is no longer responsibility of the caller, but is handled inside get_parameters immediately after final parameters are determined.
* final parameters are outputted at info log level.
* possibility to set 'global' parameters in a file `ffcx_parameters.json` in present working directory or `~/.config/ffcx`.
* the environment variable settings was removed.

This system can be added to DOLFINX for consistency once this PR is merged.

### DIFF
--- a/ffcx/parameters.py
+++ b/ffcx/parameters.py
@@ -4,6 +4,7 @@
 #
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 
+import functools
 import json
 import logging
 import os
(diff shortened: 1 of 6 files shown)


jhale:
Open question: Do we still want the environment variable stuff? If yes, perhaps it should be put in `get_parameters.py`.

michalhabera (approved):
I am happy with this.

ffcx/parameters.py
```diff
@@ -32,15 +31,79 @@
...
+
+    Notes
+    -----
+    This function sets the log level from the merged parameter values prior to
+    returning.
+
+    The ffcx_parameters.json files are cached on the first call. Subsequent
+    calls to this function use this cache.
+
+    Priority ordering of parameters is:
```
michalhabera, line 80: One detail: this should say "from highest to lowest"

PR #307, opened 2021-02-05
Interpolation

mscroggs:
- Added inverse transpose dof transformation (as this is needed in dolfinx for interpolation)
- Added RTCE, RTCF, NCE, NCF element names to the basix interface as those elements are now supported in basix


garth-wells (approved):

mscroggs:
I think we can merge this as it doesn't break anything but provides the extra functionality that interpolation needs

PR #325, opened 2021-04-08
Remove permute_dofs and unpermute_dofs functions from FFCx

mscroggs:
Added permute_dofs and unpermute_dofs functions to Basix, so these can be removed from the generated code.

This PR is coupled with FEniCS/basix#187 and FEniCS/dolfinx#1471.


michalhabera:
Could `apply_transformations_to_data` be also removed from ffcx and replace by a call to basix?

mscroggs:
Not easily, as it needs to unwrap the logic to write the transformations into generated code. I'm going to work on that in a future PR though

PR #330, opened 2021-04-16
Remove apply_dof_transformation functions

mscroggs:
These are now implemented in Basix so can be called from there directly by DOLFINx.

This PR is coupled with FEniCS/basix#193 and FEniCS/dolfinx#1489


michalhabera:
Nice. You can also remove it from element wrappers, `BaseElement`, etc?

mscroggs:
Not sure. The transformations are still written into generated code inside integral's tabulate tensor so I think they're still needed for that

michalhabera:
> Not sure. The transformations are still written into generated code inside integral's tabulate tensor so I think they're still needed for that

Right, I see.

PR #334, opened 2021-04-28
Move quadrature order increase to analysis

jorgensd:
Currently, we always add 1 to the quadrature degree of metadata.
This is not the expected behaviour.

The default quadrature schemes of order `n` in basix are exact for order `n` polynomials, and therefore we do not need to add an additional degree to the estimated polynomial.


garth-wells:
Why does 1 get added at all?

jorgensd:
> Why does 1 get added at all?

Because an exact quadrature scheme for a polynomial of degree $i$ is $i+1$.

michalhabera:
> Why does 1 get added at all?

This line seems weird, https://github.com/FEniCS/basix/blob/main/cpp/basix/quadrature.cpp#L867. It says n-point rule is used for degree 2n-2, but Gauss-Jacobi with n-point should be exact up to 2n-1, including.

michalhabera (approved):

IgorBaratta (approved):

PR #348, opened 2021-05-14
Loop transformations

chrisrichardson:
Adjusts the generated code quadrature inner loops, by:

* Using full tables (including zeros, which can be optimised out by compiler)
* Fusing loops - when all tables are the same size, this can result in huge performance gains at higher order
* Hoisting inner loop invariant float multiplications

Extensive testing has shown that all compilers perform better with these optimisations under `-Ofast -march=native`

PR #349, opened 2021-05-14
No statement for invariant only blocks

michalhabera:
Blocks with invariant only statements produced `StopIteration` exception due to #348 , because `next()` was applied on an empty iterator.

### DIFF
--- a/demo/MassDG0.ufl
+++ b/demo/MassDG0.ufl
@@ -1,0 +1,27 @@
+# Copyright (C) 2021 Igor Baratta
+#
+# This file is part of FFCx.
+#
+# FFCx is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


IgorBaratta:
I thought about that, but couldn't find a case with no statement with only "loop invariant" for rank 2.
But it’s probably safer not to assume that (as I did).

michalhabera:
> I thought about that but couldn't find a case with only "loop invariant" statements for rank 2.
> But it's probably safer not to assume that (as I did).

Assembling forms on `DG0` spaces? The generated code will look stupid with plenty optimisations for compiler, but the form should compile.

IgorBaratta (approved):

chrisrichardson:
Should there be a test that would pick this up?

PR #354, opened 2021-05-18
Simplify element table construction

chrisrichardson:
Removes some 400 lines of code, most of which was involved with creating "unique" element tables.

After removing several layers, the code is now simple, and just creates the tables for each modified terminal, checking as it goes along that they are "new tables" and not repeats of old ones.

As a consequence, since they are created in a different order, some of the table names may get changed. For example, commonly `FE12_C0_D01_Q083` and `FE12_C1_D10_Q083` may appear, where they would have both had "C0" before. But the values are as before.

By treating each subelement with its own "offset" and "stride", MixedElements will now be computed by block, rather than monolithically, so there may be a slight improvement to performance.


michalhabera:
Some description?

garth-wells (approved):

ffcx/element_interface.py
```diff
@@ -200,9 +179,32 @@ def reference_topology(self):
     def reference_geometry(self):
         return basix.geometry(self.element.cell_type)
 
-    @property
-    def is_blocked(self):
-        return False
+
+class ComponentElement(BaseElement):
+    def __init__(self, element, component):
```
garth-wells, line 184: Doc string and type hints?

ffcx/element_interface.py
```diff
@@ -223,6 +225,23 @@ def tabulate(self, nderivs, points):
             tables.append(new_table)
         return tables
 
+    def get_component_element(self, flat_component):
```
garth-wells, line 228: Doc string and type hint.

ffcx/ir/elementtables.py
```diff
@@ -407,13 +259,14 @@ def permute_quadrature_quadrilateral(points, reflections=0, rotations=0):
...
 
 
-def build_element_tables(quadrature_rule,
-                         cell,
-                         integral_type,
-                         entitytype,
-                         modified_terminals,
-                         rtol=default_rtol,
-                         atol=default_atol):
+def build_optimized_tables(quadrature_rule,
```
garth-wells, line 262: Could type hints be added?

garth-wells (commented):
Could some doc strings and type hints be added?

mscroggs:
I'm just adding some (much needed) doc strings to `element_interface.py`

PR #358, opened 2021-05-21
Fix geometry array width at 3 in generated code

### LINKED ISSUE #152: Make coordinate_dofs always 3D
`tabulate_tensor` expects the cell coordinate dofs in the same gdim as the cell is defined.
In the dolfin assembler, we need to clip the 3D geometry down to the gdim expected by ffc.

- Change ffc to use 3D geometry only.

garth-wells:
This permits a number of optimisation on the DOLFINx side when copying mesh geometry data.

Fixes #152.

### DIFF
--- a/.github/workflows/dolfin-tests.yml
+++ b/.github/workflows/dolfin-tests.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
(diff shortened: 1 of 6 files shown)


chrisrichardson (approved):

chrisrichardson (commented):
Need to revert CI

PR #365, opened 2021-06-21
Update element interface

mscroggs:
Changes to match renaming in FEniCS/basix#240.

Fixes #366 

This PR requires corresponding changes in DOLFINx (FEniCS/dolfinx#1582)

PR #367, opened 2021-06-22
Entity DOFs for quadrature element

mscroggs:
These changes follow on from #365, but I'm less sure about these.

These changes add entity dofs and entity closure dofs to `ufc_finite_element` (for quadrature elements only).

This allows FEniCS/dolfinx#1585 to be done.

This PR is into my other PR so we can either merge this with those changes, or merge that and rework this afterwards.

### DIFF
--- a/.github/workflows/dolfin-tests.yml
+++ b/.github/workflows/dolfin-tests.yml
@@ -58,7 +58,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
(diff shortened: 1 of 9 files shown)


mscroggs:
An alternative to doing it this way would be to write both entity dofs and entity closure dofs into the `ufc_dofmap`, then dolfinx can get them both from there instead of from the FiniteElement.

mscroggs:
Closing this. The more I think about it, the simpler the alternative using `ufc_dofmap` seems

PR #369, opened 2021-06-23
Add entity closure dofs to `ufc_dofmap`

mscroggs:
This allows FEniCS/dolfinx#1586 to be done

### DIFF
--- a/ffcx/codegeneration/dofmap_template.py
+++ b/ffcx/codegeneration/dofmap_template.py
@@ -17,6 +17,11 @@ void tabulate_entity_dofs_{factory_name}(int* restrict dofs, int d, int i)
 {tabulate_entity_dofs}
 }}
 
+void tabulate_entity_closure_dofs_{factory_name}(int* restrict dofs, int d, int i)
+{{
+{tabulate_entity_closure_dofs}
+}}
+
 ufc_dofmap {factory_name} =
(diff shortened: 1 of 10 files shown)


ffcx/codegeneration/dofmap_template.py
```diff
@@ -28,6 +33,11 @@
   .num_entity_dofs[2] = {num_entity_dofs[2]},
   .num_entity_dofs[3] = {num_entity_dofs[3]},
   .tabulate_entity_dofs = tabulate_entity_dofs_{factory_name},
+  .num_entity_closure_dofs[0] = {num_entity_closure_dofs[0]},
+  .num_entity_closure_dofs[1] = {num_entity_closure_dofs[1]},
+  .num_entity_closure_dofs[2] = {num_entity_closure_dofs[2]},
+  .num_entity_closure_dofs[3] = {num_entity_closure_dofs[3]},
+  .tabulate_entity_closure_dofs = tabulate_entity_closure_dofs_{factory_name},
```
chrisrichardson, line 40: Apparently, this is non-standard (a C99 extension) - it will warn you, if you try to compile with a C++ compiler.
Maybe we should be consistent, and use the same method as we are using for e.g. `value_shape` - the same is true for `num_entity_dofs` .

PR #370, opened 2021-06-28
SYCL defines

chrisrichardson:
* Allow ffcx to produce sycl compatible output


garth-wells:
Can you add some more explanation?

IgorBaratta:
> Can you add some more explanation?

Currently it produces some definitions such as:
```c++
#define tabulate_cell_a tabulate_tensor_integral_0cc1940dcb4d02004a8f4689364c2ebf7bc61339
#define dim_cell_a0 3
#define dim_cell_a1 3

#define tabulate_cell_L tabulate_tensor_integral_7d5e3ad2bad43b64846403d43f3f8ac26aa57e73
#define dim_cell_L0 3
```

So one can get the shape of the tensors at compile time.

garth-wells:
This seems a bit ad-hoc because quite a bit more required to use SYCL. How about an option that uses C++ and places the kernel in a header file, and uses `constexpr` functions to get the dims?

michalhabera:
Why is it needed to `#define` some `int`? Placing it into a member variable of our static struct wouldn't work with SYCL?

garth-wells:
@michalhabera Some values need to be fixed at runtime, e.g.might be required as template arguments.

Issue #373, opened 2021-04-08
`inner(u, div(w)) * dx` incorrect for non-affine geometric mappings


jpdean:
The following test assembles the matrix corresponding to `inner(u, div(w)) * dx` on a mesh of two square elements and two trapezium shaped elements, where `u` is from a `"DQ"` space and `w` is from an `"RTCF"` space. Due to the properties of the Piola transform, these matrices should be equal, however, the test fails.

I believe this issue is causing the vector unknown in a mixed Poisson problem to not converge in the L^2-norm (at all) on general quadrilateral meshes. There are, however, no issues on meshes of affine quadrilaterals. Note that this issue is not related to https://github.com/FEniCS/dolfinx/issues/1455.

```
import pytest
import dolfinx
import ufl
import numpy as np
from mpi4py import MPI


def assemble_matrix(k, offset):
    mesh = create_mesh(offset)

    V = dolfinx.FunctionSpace(mesh, ("DQ", k))
    W = dolfinx.FunctionSpace(mesh, ("RTCF", k + 1))

    u = ufl.TrialFunction(V)
    w = ufl.TestFunction(W)

    a = ufl.inner(u, ufl.div(w)) * ufl.dx
    A = dolfinx.fem.assemble_matrix(a)
    A.assemble()
    return A[:, :]


def create_mesh(offset):
    """Creates a mesh of two square elements if offset = 0, or two
    trapezium elements if |offset| > 0.
    """
    x = np.array([[0, 0],
                  [1, 0],
                  [0, 0.5 + offset],
                  [1, 0.5 - offset],
                  [0, 1],
                  [1, 1]])
    cells = np.array([[0, 1, 2, 3],
                      [2, 3, 4, 5]])

    ufl_mesh = ufl.Mesh(ufl.VectorElement("Lagrange", "quadrilateral", 1))
    mesh = dolfinx.mesh.create_mesh(MPI.COMM_WORLD, cells, x, ufl_mesh)
    return mesh


@pytest.mark.parametrize("k", [0, 1, 2])
def test_general_quads(k):
    # Assemble matrix on a mesh of square elements
    A_square = assemble_matrix(k, 0)
    # Assemble matrix on a mesh of trapezium elements
    A_trap = assemble_matrix(k, 0.25)

    # Due to the properties of the Piola transform, A_square and A_trap
    # should be equal
    assert np.allclose(A_square, A_trap, atol=1e-6)
```

jpdean:
[...]

PR #374, opened 2021-07-04
Prevent global Jacobian derivatives being interpreted as local derivatives

jpdean:
This PR stops global Jacobian derivatives being interpreted as local derivatives and fixes https://github.com/FEniCS/ffcx/issues/373. An alternative approach to removing  `preserve_geometry_types=(ufl.classes.Jacobian, ),` would be to modify UFL such that `preprocessed_form` contains local Jacobian derivatives instead of global derivatives, as described here https://github.com/FEniCS/ufl/issues/62.


jpdean:
Closing this because it can (and I think probably should) be fixed in UFL

PR #376, opened 2021-08-02
 Update Basix interface to allow variants

mscroggs:
This is coupled with FEniCS/dolfinx#1651 and FEniCS/basix#252.


ffcx/codegeneration/access.py
```diff
@@ -261,7 +261,7 @@ def cell_vertices(self, e, mt, tabledata, num_points):
         ufl_scalar_element, = set(coordinate_element.sub_elements())
         assert ufl_scalar_element.family() in ("Lagrange", "Q", "S")
 
-        basix_scalar_element = create_element(ufl_scalar_element)
+        basix_scalar_element = create_element(ufl_scalar_element, "equispaced")
```
garth-wells, line 264: Could we make the lattice type an enum? (defined in Basix)

ffcx/element_interface.py
```diff
@@ -25,30 +25,37 @@
 }
 
 
-def create_element(ufl_element):
+def create_element(ufl_element, *args):
```
garth-wells, line 28: I've always had a dislike of `*args` for readability reasons. Is there a simple way around using it? If no, we need a detailed docstring.

ffcx/ir/elementtables.py
```diff
@@ -54,7 +54,7 @@ def clamp_table_small_numbers(table,
 
 
 def get_ffcx_table_values(points, cell, integral_type, ufl_element, avg, entitytype,
-                          derivative_counts, flat_component):
+                          derivative_counts, flat_component, finite_element_args):
```
garth-wells, line 57: Add some docs - not sure what `finite_element_args` is.

garth-wells (changes requested):

ffcx/ir/elementtables.py
mscroggs: Thankfully, `finite_element_args` is no longer necessary so the code is now a little simpler

ffcx/codegeneration/ufc.h
```diff
@@ -123,6 +123,19 @@ extern "C"
...
 
+    /// Return the Basix identifier of the family of the finite element function space
+    int basix_family;
+
+    /// Return the Basix identifier of the cell shape
+    int basix_cell;
+
+    /// Set to true if Basix's create_element function for this element
+    /// needs to be given a lattice type
+    bool needs_lattice_type;
```
garth-wells, line 134: Have about removing this and using `lattice_type = -1` if the lattice isn't needed?

garth-wells (approved):

PR #377, opened 2021-08-06
Pad loops, not only tables

IgorBaratta:
We have the option of padding arrays:
```
--padlen PADLEN       Pads every declared array in tabulation kernel such that its last dimension is divisible by given value. (default=1)
```

But in general that's not really helpful if we don't adjuts the loop bounds.
The idea of this PR is to adjust these bounds, in special inner loops that can be vectorized.

For two dimensional structures [m*n] pad only the  leading dimension: 
```
p = m + (padlen - m%padlen)%padlen
A[n*(m+p)]
```

### DIFF
--- a/demo/Poisson.ufl
+++ b/demo/Poisson.ufl
@@ -20,7 +20,7 @@
 #
 # Compile this form with FFCx: ffcx Poisson.ufl
 
-element = FiniteElement("Lagrange", triangle, 1)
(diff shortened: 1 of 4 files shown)


garth-wells:
What impacts can this have on performance?

IgorBaratta:
> What impacts can this have on performance?

The impact can range from none to 40% speedup for rank-1 tensors, depending on the compiler and the complexity of the form.

PR #378, opened 2021-08-09
Revert branches to main

mscroggs:


michalhabera:
Was this needed? I thought I disabled cross-repo status check as required.

mscroggs:
Don't we want the cross repo check to be required so we can't merge things that break DOLFINx without noticing?

michalhabera:
So my idea was: Develop your cross repo PR as usual, dolfinx/my-feature and ffcx/my-feature. Their CIs points to each other, once both PRs are green and approved, just for the merging action you'd set ffcx/my-feature CI point to dolfinx/main. This CI would fail, but you can merge it. Then you'd point dolfinx/my-feature to ffcx/main, which would not fail, and you can merge it.

mscroggs:
Ah I see, so we can reset the branch names before merging the first PR now, then there's no need for a second PR. I'll do that from now on

PR #380, opened 2021-08-10
Add diagonal mass matrix demo

adeebkor:


michalhabera:
Can you comment why "diagonal" table is not treated as a table type?

IgorBaratta:
> Can you comment why "diagonal" table is not treated as a table type?

We are just experimenting.
We just wanted to modify less code until we know that it's good to have a new table type.

btw, it was a PR to a branch, not to main.

michalhabera:
> > Can you comment why "diagonal" table is not treated as a table type?
> 
> We are just experimenting.
> We just wanted to modify less code until we know that it's good to have a new table type.
> 
> btw, it was a PR to a branch, not to main.

Ah, I did not notice that :) In the past there was `ttype="identity"` for this purpose I think, have a look at the code at that time.

PR #382, opened 2021-08-16
Discontinuous elements

mscroggs:
Use the `discontinuous` input added in FEniCS/basix#260.

This PR is coupled with FEniCS/dolfinx#1670 and FEniCS/basix#260.

### DIFF
--- a/ffcx/codegeneration/finite_element.py
+++ b/ffcx/codegeneration/finite_element.py
@@ -41,7 +41,7 @@ def generator(ir, parameters):
     d["family"] = f"\"{ir.family}\""
     d["num_sub_elements"] = ir.num_sub_elements
     d["block_size"] = ir.block_size
-    d["discontinuous"] = 1 if ir.discontinuous else 0
+    d["discontinuous"] = "true" if ir.discontinuous else "false"
 
     if ir.lattice_type is None:
         d["lattice_type"] = -1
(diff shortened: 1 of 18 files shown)


ffcx/element_interface.py
```diff
@@ -316,6 +335,11 @@ def cell_type(self):
         """Get the Basix cell type used to initialise the element."""
         return self._cell
 
+    @property
+    def discontinuous(self):
+        """Get a bool indicating whether the discontinuous version of the element is used."""
```
garth-wells, line 340: Rather than putting `bool` in the docstring, could we make it a return type hint?

And should the function go into `BaseElement`?

ffcx/element_interface.py
```diff
@@ -385,6 +409,11 @@ def cell_type(self):
         """Get the Basix cell type used to initialise the element."""
         return self.element.cell_type
 
+    @property
+    def discontinuous(self):
```
garth-wells, line 413: See comment above on type hints.

ffcx/codegeneration/finite_element.py
```diff
@@ -41,6 +41,7 @@ def generator(ir, parameters):
     d["family"] = f"\"{ir.family}\""
     d["num_sub_elements"] = ir.num_sub_elements
     d["block_size"] = ir.block_size
+    d["discontinuous"] = 1 if ir.discontinuous else 0
```
garth-wells, line 44: Could we use `true` and `false` here rather than `1` and `0`?

garth-wells (changes requested):

garth-wells (approved):

PR #389, opened 2021-09-20
Replace lattice type with Lagrange variant

mscroggs:
This is necessary following FEniCS/basix#279


mscroggs:
CI is currently failing as coveralls is down for maintenance. Tests need to be rerun once its back up

PR #409, opened 2021-10-28
Store coordinate element in ufc_form

### LINKED ISSUE #406: Add geometry element to ufc_form
We can then check in DOLFINx that the correct element is used in CoordinateElement

mscroggs:
Fixes #406.

### DIFF
--- a/demo/test_demos.py
+++ b/demo/test_demos.py
@@ -23,8 +23,9 @@ def test_demo(file):
         # Skip demos that use elements not yet implemented in basix
         pytest.skip()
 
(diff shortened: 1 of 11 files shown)


garth-wells (commented):
Should some tests be added?

michalhabera:
I think it would make more sense to store it in `ufc_integral`. There are several integrals per form and each integral could have domain of different coordinate element - especially towards transition to mixed topology. Also, shouldn't ufc_dofmap be attached too (maybe some form of "function space")?

michalhabera (approved):

PR #416, opened 2021-11-03
Check for unused variables in generated code

IgorBaratta:


michalhabera:
There still seems to be a number of unused zeros tables, I'll have a look.

PR #418, opened 2021-11-13
Update existing tables

IgorBaratta:
The "existing tables" list is not being updated, which causes repeated tables to be created.


michalhabera:
> The "existing tables" list is not being updated, which causes repeated tables to be created.

Existing unique tables are meant to be updated outside this function, in https://github.com/FEniCS/ffcx/blob/main/ffcx/ir/integral.py#L253 I changed`build_optimized_tables` function so it does not modify its arguments.

michalhabera (approved):

PR #425, opened 2021-11-21
Use XDG_CONFIG_HOME when searching for config file

matsievskiysv:
Use `XDG_CONFIG_HOME` when searching for `ffcx_parameters.json` configuration file. Edit doc line to match <https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/jit.py> style.
Related PR in dolfinx: https://github.com/FEniCS/dolfinx/pull/1822

### DIFF
--- a/ffcx/parameters.py
+++ b/ffcx/parameters.py
@@ -9,7 +9,7 @@ import json
 import logging
 import os
 import os.path
-import pathlib
+from pathlib import Path
 import pprint
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


jhale:
Great. If you could use the default kwarg to os.getenv to set the default as in the XDG spec then I'd be happy to accept the PR. Same applies to the DOLFINx PR.

matsievskiysv:
@jhale Done

PR #427, opened 2021-11-26
Enable static type checking with mypy

chrisrichardson:
Adds type annotation where needed for static type checking with `mypy`, which may pick up some bugs, or omissions.


garth-wells:
@chrisrichardson can you add some text to the description on what this does?

IgorBaratta (approved):

garth-wells (approved):

PR #431, opened 2021-11-29
Remove ufc_blocked_element enum

### LINKED ISSUE #417: Remove `ufc_blocked_element` enum
Ambiguous what this really means, and we have the block size to get the block data.

garth-wells:
Fixes #417.


mscroggs (approved):
I changed what `element_type` returns, as a user could create a vectorelement of a quadrature element

PR #432, opened 2021-12-01
Update and extend ufc_expression

michalhabera:
This PR updates signature for `ufc_expression` tabulation kernel to use `ufc_tabulate_tensor_float32`, `ufc_tabulate_tensor_float64`, etc. (shared signature with `ufc_form`).

More information is added to ufc_expression to allow rank-1 expressions in dolfinx, i.e. Expressions containing UFL Argument. This is useful to build point-wise operators, see dolfinx PR.


jorgensd:
Isn't this PR missing `apply_dof_transformations` for the more exotic finite elements (N1 curl, N1 div)?

michalhabera:
> Isn't this PR missing `apply_dof_transformations` for the more exotic finite elements (N1 curl, N1 div)?

This is responsibility of kernel caller, right? My understanding is dof transformations would be called in dolfinx after tabulation kernel. But I did not add any example which would `needs_facet_permutations == True` into dolfinx tests yet.

jorgensd:
> > Isn't this PR missing `apply_dof_transformations` for the more exotic finite elements (N1 curl, N1 div)?
> 
> This is responsibility of kernel caller, right? My understanding is dof transformations would be called in dolfinx after tabulation kernel. But I did not add any example which would `needs_facet_permutations == True` into dolfinx tests yet.

Sorry, I put this comment on the wrong PR, was meant for the dolfinx one.

PR #437, opened 2022-01-11
Rename `ufc.h` -> `ufcx.h` and add CMake files for header-only install

garth-wells:
Rename `ufc.h` to `ufcx.h` to distinguish it from the old `ufc.h`.

Adds CMake files for the optional installation of `ufcx.h` without installing the whole `ffcx` Python module. This functionality is not expected to be needed by users, but will help with packing. The reason is `ufcx.h` is independent of the Python and therefore should be packaged separately. 

Will require updates to DOLFINx (https://github.com/FEniCS/dolfinx/pull/1924).


cmake/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 3.16)
+
+project(ufcx VERSION 0.3.1 DESCRIPTION "UFCx interface header for finite element kern"
```
michalhabera, line 3: Is this sentence cut on purpose?

michalhabera (approved):

.github/workflows/dolfin-tests.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         with:
           path: ./dolfinx
           repository: FEniCS/dolfinx
-          ref: main
+          ref: ufcx
```
jhale, line 48: Remember to remove.

jhale (approved):

PR #438, opened 2022-01-12
Zero table elimination for non-scalar graphs

jorgensd:
Resolves https://github.com/FEniCS/dolfinx/issues/1926


ffcx/ir/integral.py
```diff
@@ -95,11 +96,11 @@ def compute_integral_ir(cell, integral_type, entitytype, integrands, argument_sh
         tables = {v.name: v.values for v in mt_table_reference.values()}
 
         S_targets = [i for i, v in S.nodes.items() if v.get('target', False)]
+        num_targets = numpy.int32(numpy.prod(expression.ufl_shape))
+        assert(num_targets == len(S_targets))
```
michalhabera, line 100: This is not true if there is a root node, which contributes to multiple components -- consider a symmetric tensor. Similarly, you are taking just 1st component below, which is wrong.

jorgensd: I've tried to fix this now. What do you think?

jorgensd: Test added in: https://github.com/FEniCS/dolfinx/pull/1928/commits/51405631f78a056d056b30e7d445200e96b32878

ffcx/ir/integral.py
```diff
@@ -112,10 +113,22 @@ def compute_integral_ir(cell, integral_type, entitytype, integrands, argument_sh
...
+            if value_rank == 0:
+                # Rebuild scalar target expressions and graph (this may be
+                # overkill and possible to optimize away if it turns out to be
+                # costly)
+                expression = S.nodes[S_targets[0]]['expression']
+            elif value_rank == 1 or value_rank == 2:
+                expressions = [None, ] * num_components
+                for target in S_targets:
+                    for comp in S.nodes[target]["component"]:
+                        expressions[comp] = S.nodes[target]["expression"]
```
michalhabera, line 125: Can you add assert here, that this is not overwriting some component added earlier, `assert(expressions[comp] == None)` before. This should never be the case, just to be safe.

ffcx/ir/integral.py
```diff
@@ -112,10 +113,22 @@ def compute_integral_ir(cell, integral_type, entitytype, integrands, argument_sh
...
-            # Rebuild scalar target expressions and graph (this may be
-            # overkill and possible to optimize away if it turns out to be
-            # costly)
-            expression = S.nodes[S_targets[0]]['expression']
+            if value_rank == 0:
+                # Rebuild scalar target expressions and graph (this may be
+                # overkill and possible to optimize away if it turns out to be
+                # costly)
+                expression = S.nodes[S_targets[0]]['expression']
+            elif value_rank == 1 or value_rank == 2:
```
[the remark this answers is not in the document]
jorgensd: What ufl function can we use for rank>2? 
rank 1: as_vector
rank 2: as_tensor
rank >2:?

ffcx/ir/integral.py
```diff
@@ -112,10 +113,22 @@ def compute_integral_ir(cell, integral_type, entitytype, integrands, argument_sh
...
+                expressions = [None, ] * num_components
+                for target in S_targets:
+                    for comp in S.nodes[target]["component"]:
+                        expressions[comp] = S.nodes[target]["expression"]
+                if value_rank == 1:
+                    expression = ufl.as_vector(expressions)
+                else:
+                    expression = ufl.as_tensor(numpy.asarray(expressions).reshape(expression.ufl_shape))
```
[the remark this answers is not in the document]
jorgensd: It is to be able to reshape the list to a NxM tensor matching the ufl shape.

ffcx/ir/integral.py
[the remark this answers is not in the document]
michalhabera: `as_tensor` is the most general function, because mathematically, so is "tensor". UFL uses special cases, if rank==1 then as_vector, if rank==2 then as_matrix. But I think you should be able all cases with as_tensor. It also suggests you should not use numpy's matrix below.

ffcx/ir/integral.py
```diff
@@ -112,10 +115,18 @@ def compute_integral_ir(cell, integral_type, entitytype, integrands, argument_sh
                 if deps:
                     v['expression'] = v['expression']._ufl_expr_reconstruct_(*deps)
 
-            # Rebuild scalar target expressions and graph (this may be
-            # overkill and possible to optimize away if it turns out to be
-            # costly)
-            expression = S.nodes[S_targets[0]]['expression']
+            if value_rank == 0:
```
michalhabera, line 118: This should work through general codepath below, so no need for this branch.

test/test_jit_expression.py
```diff
@@ -141,3 +142,162 @@ def test_rank1(compile_args):
     u_correct = np.array([f[1], f[0]]) + gradf0
 
     assert np.allclose(u_ffcx, u_correct.T)
+
+
+def test_elimiate_zero_tables_scalar(compile_args):
```
michalhabera, line 147: As all will fo through same codepath, I think it suffices to have the tensor test.

michalhabera (approved):

test/test_jit_expression.py
```diff
@@ -141,3 +142,63 @@ def test_rank1(compile_args):
     u_correct = np.array([f[1], f[0]]) + gradf0
 
     assert np.allclose(u_ffcx, u_correct.T)
+
+
+def test_elimiate_zero_tables_tensor(compile_args):
```
garth-wells, line 147: Not clear how this is being 'tested'. Does it rely on struct compiler flags picking up the issue?

test/test_jit_expression.py
```diff
@@ -141,3 +142,63 @@ def test_rank1(compile_args):
...
+    # Get vectices of cell
+    # Coords storage XYZXYZXYZ
+    basix_c_e = basix.create_element(basix.ElementFamily.P, basix.cell.string_to_type(cell), 1, False)
+    coords = basix_c_e.points
+
+    # Using same basix element for coordinate element and coefficient
+    coeff_points = basix_c_e.points
+
+    # Compile expression
+    # Get interpolation points of CG 2
```
garth-wells, line 171: Use 'Lagrange' rather than 'CG'

test/test_jit_expression.py
```diff
@@ -141,3 +142,63 @@ def test_rank1(compile_args):
...
+    obj, module, code = ffcx.codegeneration.jit.compile_expressions(
+        [(expr, points)], cffi_extra_compile_args=compile_args)
+
+    ffi = cffi.FFI()
+    expression = obj[0]
+    c_type, np_type = float_to_type("double")
+
+    output = np.zeros(9 * points.shape[0], dtype=np_type)
+
+    def u_expr(x):
```
garth-wells, line 183: Why is this function required?

test/test_jit_expression.py
jorgensd: It is not needed. I'll remove it

michalhabera: What is not needed? All tests which call JIT take compile_args fixture. Thats how we control JIT compilation parameters, especially it uses "-O1" flag to speed them up.

test/test_jit_expression.py
garth-wells:
```
u_coeffs = coeff_points.T[0] + 2 * coeff_points.T[1] 
```
?

garth-wells (approved):

test/test_jit_expression.py
```diff
@@ -141,3 +142,59 @@ def test_rank1(compile_args):
     u_correct = np.array([f[1], f[0]]) + gradf0
 
     assert np.allclose(u_ffcx, u_correct.T)
+
+
+def test_elimiate_zero_tables_tensor():
```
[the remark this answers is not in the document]
jorgensd: Because @garth-wells  was wondering if they were necessary for the test to run, which they are not.
If you have a good reason for having them here, I will add them back in.

PR #419, opened 2021-11-14
Improve data access for coefficients

IgorBaratta:
Map linear stride to unit stride in coefficients.

The idea is to try to avoid linear access (non-unit stride) within the quadrature loop.

Original code:
```c++
    for (int iq = 0; iq < 4; ++iq)
    {
        // Quadrature loop body setup for quadrature rule fad
        // Varying computations for quadrature rule fad
        double w0_d100_c0 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d100_c0 += w[ic * 3] * FE18_C0_D100_Qfad[0][0][iq][ic];
        double w0_d010_c0 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d010_c0 += w[ic * 3] * FE18_C0_D010_Qfad[0][0][iq][ic];
        double w0_d001_c0 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d001_c0 += w[ic * 3] * FE18_C0_D001_Qfad[0][0][iq][ic];
        double w0_d100_c1 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d100_c1 += w[ic * 3 + 1] * FE18_C0_D100_Qfad[0][0][iq][ic];
        double w0_d010_c1 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d010_c1 += w[ic * 3 + 1] * FE18_C0_D010_Qfad[0][0][iq][ic];
        double w0_d001_c1 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d001_c1 += w[ic * 3 + 1] * FE18_C0_D001_Qfad[0][0][iq][ic];
        double w0_d100_c2 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d100_c2 += w[ic * 3 + 2] * FE18_C0_D100_Qfad[0][0][iq][ic];
        double w0_d010_c2 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d010_c2 += w[ic * 3 + 2] * FE18_C0_D010_Qfad[0][0][iq][ic];
        double w0_d001_c2 = 0.0;
        for (int ic = 0; ic < 10; ++ic)
            w0_d001_c2 += w[ic * 3 + 2] * FE18_C0_D001_Qfad[0][0][iq][ic];ic];
```

New code:
[...]

PR #446, opened 2022-01-23
Merge pre-definitions across quadrature loops

michalhabera:
Since https://github.com/FEniCS/ffcx/pull/419 blocked coefficients have a quadrature loop independent code to allow their unit-stride access. These "pre-definitions" must be merged across quadrature loops since they lead to name clash and compiler error otherwise. 

This PR solves the above bug, it also improves pre-definitions re-use in some cases.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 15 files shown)


IgorBaratta:
Looks good to me.
But could you add a test (ufl file) that would catch this bug?

michalhabera:
> Looks good to me. But could you add a test (ufl file) that would catch this bug?

Extended existing UFL file (would fail on main).

PR #450, opened 2022-02-02
Set pygraphviz version to 1.7

### LINKED ISSUE #449: `PyGraphviz requires Python version 3.8 or later (3.7 detected).`
I'm getting this error when running CI (https://github.com/FEniCS/ffcx/runs/5035402568?check_suite_focus=true). Do we need to bump our minimum Python requirement to 3.8?

mscroggs:
Fixes #449

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ install_requires =
 [options.extras_require]
 docs = sphinx; sphinx_rtd_theme
 lint = flake8; pydocstyle[toml]
-optional = pygraphviz == 1.7
+optional = pygraphviz >= 1.7
 test = pytest >= 6.0; sympy
 ci =
     coverage
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,7 +51,7 @@ install_requires =
 [options.extras_require]
 docs = sphinx; sphinx_rtd_theme
 lint = flake8; pydocstyle[toml]
-optional = pygraphviz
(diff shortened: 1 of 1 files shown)


garth-wells:
I don't think this is a sustainable fix. I'd be happier to drop Python 3.7 from the CI, and have 3.8. 3.9 and 3.10.

jorgensd:
> I don't think this is a sustainable fix. I'd be happier to drop Python 3.7 from the CI, and have 3.8. 3.9 and 3.10.

I think it is quite strict to drop Python 3.7 before it reaches end of life (June 2023).

garth-wells:
> > I don't think this is a sustainable fix. I'd be happier to drop Python 3.7 from the CI, and have 3.8. 3.9 and 3.10.
> 
> I think it is quite strict to drop Python 3.7 before it reaches end of life (June 2023).

We're not dropping Python 3.7. It's about the CI. FFCx has a tiny dependency on `pygraphviz` - the functionality it provides is peripheral. A point is that we have always strived to work with the the latest releases of libraries. Deviating form this leads to major dependency management challenges.

jhale:
Greater than or equal to is more flexible for dependencies.

jhale:
Sorry I misread the issue.

I am going to close this this. I agree with @garth-wells, we should drop 3.7 from the CI. We don't have the resources to deal with being compatible with legacy packages.

PR #453, opened 2022-02-03
Improve expression C++ interface

jorgensd:
Extend expression interface to allow `ufl.FiniteElementBase` as input arguments (except MixedSpaces).
This avoid manual definitions of interpolation points, which in turn can be used to compute secondary variables without projections.

A typical `.ufl` file would look like:
```python
cell = triangle
v_el = VectorElement("Lagrange", cell, 1)
mesh = Mesh(v_el)

el = FiniteElement("CG", cell, 2)
el_int = VectorElement("Discontinuous Lagrange", cell, 1)
me = MixedElement([el, el_int])
V = FunctionSpace(mesh, me)
u = Coefficient(V)

du0 = grad(u[0])
du1 = grad(u[1])

q_el = FiniteElement("Quadrature", cell, 3, quad_scheme="default")
Q = FunctionSpace(mesh, q_el)
q = Coefficient(Q)
powq = 3*q**2

expressions = [(du0, el), (du1, el_int), (powq, q_el)]
```


ffcx/ir/representation.py
```diff
@@ -471,7 +471,22 @@ def _compute_expression_ir(expression, index, prefix, analysis, parameters, visu
...
+        if isinstance(point_element, BlockedElement):
+            points = point_element.sub_element.element.points
+        elif isinstance(point_element, QuadratureElement):
+            points = point_element._points
+        elif isinstance(point_element, MixedElement):
+            raise ValueError("Evaluation of an expression with a mixed element is not allowed")
+        else:
+            points = point_element.element.points
+    else:
+        points = expression[1]
```
jorgensd, line 488:
```suggestion
        points = numpy.asarray(expression[1])
```

ffcx/ir/representation.py
```diff
@@ -471,7 +471,22 @@ def _compute_expression_ir(expression, index, prefix, analysis, parameters, visu
     ir["name"] = naming.expression_name(original_expression, prefix)
 
     original_expression = expression[2]
-    points = expression[1]
+
+    # If the second element argument of the Expression tuple is an ufl.FiniteElement,
```
jorgensd, line 475:
```suggestion
    # If the second element argument of the Expression tuple is defined as a finite element from ufl,
    # get the corresponding interpolation points
```

jhale (approved):

ffcx/analysis.py
```diff
@@ -67,7 +67,10 @@ def analyze_ufl_objects(ufl_objects: typing.List,
             coordinate_elements += [ufl_object.ufl_coordinate_element()]
         elif isinstance(ufl_object[0], ufl.core.expr.Expr):
             original_expression = ufl_object[0]
-            points = numpy.asarray(ufl_object[1])
+            if isinstance(ufl_object[1], ufl.FiniteElementBase):
```
michalhabera, line 70: I would suggest you create basix element and fetch points at this point. It is more error-prone if we keep this generic tuple for too long in the pipeline.

ffcx/ir/representation.py
```diff
@@ -471,7 +471,23 @@ def _compute_expression_ir(expression, index, prefix, analysis, parameters, visu
     ir["name"] = naming.expression_name(original_expression, prefix)
 
     original_expression = expression[2]
-    points = expression[1]
+
+    # If the second element argument of the Expression tuple is defined as a finite element from ufl,
+    # get the corresponding interpolation points
+    # get the interpolation points
+    if isinstance(expression[1], ufl.FiniteElementBase):
```
michalhabera, line 478: This will be simpler if you move it to analysis.py (see comment above).

michalhabera (changes requested):

michalhabera (approved):

jorgensd:
Changes suggested by @michalhabera  has not implemented as it breaks the hashing of the expression.

michalhabera:
I am happy with this, but in longer terms we'll need to formalise notion of "expression with evaluation points", so tuple of multivalued type isn't entering the interface. Any comments @garth-wells ?

garth-wells:
> I am happy with this, but in longer terms we'll need to formalise notion of "expression with evaluation points", so tuple of multivalued type isn't entering the interface. Any comments @garth-wells ?

The line
```
expressions = [(du0, el), (du1, el_int), (powq, q_el)]
```
looks cryptic to me. Could `namedtuple` or something else make it clearer?

jorgensd:
> `namedtuple`

Im not sure how namedtuple would help here. It is simply a list of tuples (ufl_expression, ufl_element), where we want to generate code for evaluating the `ufl_expression` at the interpolation points of the `ufl_element`.

garth-wells:
@jorgensd now that we're using 'plain' .py files, could you try getting the same functionality by importing a Basix element? It wouldn't be a completely smooth interface just yet, but I think the more programmatic approach will serve us better going forward.

jorgensd:
Superseded by #463

PR #459, opened 2022-02-17
Make UFL files Python

michalhabera:
Along with https://github.com/FEniCS/ufl/pull/90


.github/workflows/pythonapp.yml
```diff
@@ -82,9 +82,12 @@ jobs:
         # when there are test failures
         if: always()
 
-      - name: Runs demos
+      - name: Run FFCx and UFL demos
         run: |
-          python -m pytest demo/test_demos.py
+          git clone --branch michal/py-form-files https://github.com/FEniCS/ufl.git
```
garth-wells, line 87: There are 'actions' form cloning other repos, see https://github.com/FEniCS/basix/blob/main/.github/workflows/dolfin-tests.yml#L46 for an example.

demo/Constant.py
```diff
@@ -1,32 +1,35 @@
 # Copyright (C) 2007 Anders Logg
 #
-# This file is part of FFCx.
+# This file is part of UFL.
```
garth-wells, line 3: Is this change correct (and other similar changes)?

garth-wells (changes requested):
Nice to get this changed. A few minor comments inline.

PR #470, opened 2022-04-01
Add support for custom Basix elements

mscroggs:
This is coupled with https://github.com/FEniCS/basix/pull/405 and https://github.com/FEniCS/dolfinx/pull/2113.

### DIFF
--- a/ffcx/codegeneration/ufcx.h
+++ b/ffcx/codegeneration/ufcx.h
@@ -64,11 +64,13 @@ extern "C"
     ufcx_basix_element = 0,
     ufcx_mixed_element = 1,
     ufcx_quadrature_element = 2,
-    ufcx_custom_element = 3
+    ufcx_custom_element = 3,
+    ufcx_basix_custom_element = 4
   } ufcx_element_type;
 
   /// Forward declarations
   typedef struct ufcx_finite_element ufcx_finite_element;
+  typedef struct ufcx_basix_custom_finite_element ufcx_basix_custom_finite_element;
   typedef struct ufcx_dofmap ufcx_dofmap;
   typedef struct ufcx_function_space ufcx_function_space;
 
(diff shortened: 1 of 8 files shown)


ffcx/codegeneration/ufcx.h
```diff
@@ -64,11 +64,13 @@ extern "C"
     ufcx_basix_element = 0,
     ufcx_mixed_element = 1,
     ufcx_quadrature_element = 2,
-    ufcx_custom_element = 3
+    ufcx_custom_element = 3,
```
michalhabera, line 67: What is the difference between custom element and basix custom element?

mscroggs: Good question...

mscroggs: I've just checked and ufcx_custom_element isn't used anywhere, so I've removed it.

(As it was there before this PR, I wrongly assumed it was being used for something else)

ffcx/codegeneration/ufcx.h
```diff
@@ -144,8 +145,67 @@ extern "C"
...
+    int* value_shape;
+
+    /// The number of rows in the wcoeffs matrix
+    int wcoeffs_rows;
+
+    /// The number of columnss in the wcoeffs matrix
+    int wcoeffs_cols;
+
+    /// The data contained in wcoeffs
+    double* wcoeffs;
```
michalhabera, line 173: I think UFCx.h should be mostly self-contained. This doesn't really help in explaining what `wcoeffs` stands for. I understand this is for basix element, so the meaning comes from there, but few more words with possible pointer where to look at would be nice.

mscroggs: I've added a short description. I'll also add a link to more info once https://github.com/FEniCS/basix/issues/404 is done

PR #472, opened 2022-04-12
Cache Basix elements to avoid computing basis functions multiple times

mscroggs:
I put a print in the `basix::FiniteElement` initialisation. When compiling a form, each element was being created multiple times:

image

After this PR, each element is only computed twice (once by FFCx, and once by DOLFINx):

image

For higher degree elements, this will speed up compilation of forms by a fair amount


michalhabera:
Could you just use `cache` deocator from `functools`?

PR #484, opened 2022-04-27
Update version number

mscroggs:
@chrisrichardson Does this one need updating too?

Seems to be causing failures like this: https://github.com/FEniCS/dolfinx/runs/6193714382?check_suite_focus=true

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.16)
 
-project(ufcx VERSION 0.3.1 DESCRIPTION "UFCx interface header for finite element kernels"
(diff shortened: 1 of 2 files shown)


ma595:
Yes, this change will also enable the fenics-dolfinx spack package to build with fenics-ufcx.

chrisrichardson:
Yes, I agree. There are version numbers in too many places.
Maybe we will need to make a 0.4.1 release of ffcx? Or I can just retag? Any thoughts @ma595?

chrisrichardson (approved):

ma595:
I slightly prefer retagging 0.4.0.

mscroggs:
I agree. Having a version 0.4.0 that's out there but broken seems a bad idea

PR #485, opened 2022-04-28
add version consistency checking test

mscroggs:
Added a test that checks that version numbers in different locations agree


jhale:
Looks ok although scripts like this can sometimes be a bit flakey.

I'm happy to merge but it is missing the (more difficult) version numbers in the `ufcx.h` header and `cmake/CMakeLists.txt`.

cmake/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.16)
 
-project(ufcx VERSION 0.4.3 DESCRIPTION "UFCx interface header for finite element kernels"
+project(ufcx VERSION 0.4.3.0 DESCRIPTION "UFCx interface header for finite element kernels"
```
mscroggs, line 3: I'm uncertain whether this change is correct or if the test needs changing

PR #502, opened 2022-06-23
Use float type that matches scalar type

garth-wells:
Previously `double` was used for basis functions and geometry. This eliminated performance benefits of using reduced precision for the fields. This change uses the same float type as for the fields.


IgorBaratta (approved):
LGTM!

PR #504, opened 2022-06-23
Do not unroll loops for jacobian computation

IgorBaratta:


chrisrichardson:
I guess this is OK, but I wonder if it makes much difference.
Which kernels will benefit from it? Surely all the time is spent in the main quadrature loop.

IgorBaratta:
> I guess this is OK, but I wonder if it makes much difference. Which kernels will benefit from it? Surely all the time is spent in the main quadrature loop.

Not much difference for simplices, but for sum factorization we need this loop structure.

garth-wells (approved):

chrisrichardson (approved):

PR #520, opened 2022-08-03
Allow `QuadratureElement` with custom quadrature rule

### LINKED ISSUE #510: Support custom Quadrature-rules in QuadratureElement
For people do be able to use quadrature elements with other schemes than those supported through basix, we would need an alternative interface taking in quadrature points/weights.

mscroggs:
Fixes #510


garth-wells:
@mscroggs could you resolve the merge conflicts?

garth-wells (approved):

Issue #290, opened 2020-11-29
increasing the entropy of the jit hash


drew-parsons:
Currently, generated jit files cause a clash under certain circumstances such as change of Python version. For instance upgrading from python3.8 to python3.9, the generated files have the same name (so a new one is not generated) but are not actually compatible so there is runtime failure.  

Discussion at https://github.com/FEniCS/dolfinx/issues/1227 addresses enabling different cache_dirs for different projects (e.g. one project using python3.8, another using python3.9).

But regardless of cache_dir, that discussion favours the idea of increase the entropy of the hash used to identify generated jit files, e.g. seeding the python version into the hash.

There could be a range of different additional parameters worth feeding into the hash. Python version in the first instance, possibly other compilation parameters.

The action would take place in [`compute_signature`](https://github.com/FEniCS/ffcx/blob/cc9064e3971695352ceb59ac64a0cfb4b011d756/ffcx/naming.py#L13) in ffcx/naming.py. The desirable additional elements would want to be added to the `signatures` list at [l.79](https://github.com/FEniCS/ffcx/blob/cc9064e3971695352ceb59ac64a0cfb4b011d756/ffcx/naming.py#L79).

michalhabera:
@drew-parsons Actually, the action needs to take place here (and several similar calls below) https://github.com/FEniCS/ffcx/blob/master/ffcx/codegeneration/jit.py#L148

It already adds extra compiler args, which is good. Do you want to add the python version in these places in a very similar way and make it into PR? More general, any state which would effect the C code generated by CFFI, thus ABI compatibility of compiled `.so` files should be added there. The hash in the filename should be equivalent to the hash of the code in `libffcx_forms_foo.c`.

drew-parsons:
Ah, right. I'll track done some useful parameters to include. Possibly PY_VERSION_HEX from /usr/include/python*/patchlevel.h.
Except we likely want patch level compatibility in the hash, so might be better with just PY_MAJOR_VERSION and PY_MINOR_VERSION.

drew-parsons:
Except jit.py is already python. So, maybe `sys.api_version` + `sys.abiflags`. Will need to look into whether that's better than `sys.version_info`.

michalhabera:
However, in my installation `sys.api_version` returns 1013 for all python version I have (python2.7, python3.8, python3.9).

[...]

PR #527, opened 2022-08-22
Add Python's CFLAGS, SOABI to cache key

minrk:
SOABI is _often_ enough, but doesn't capture arm vs intel on mac (`sysconfig.get_platform()` does, but also includes the _current_ OS version rather than the compile target - invalidating the cache on annual upgrades). CFLAGS does capture target architecture, base SDK, etc.

Improves the situation for #290 (not sure if it's enough to close)

Avoids cache conflicts for:

- different Python versions, debug builds (SOABI)
- different architectures on mac (CFLAGS)
- different base SDKs, possible other compilation inputs

CFLAGS _will_ result in some potentially overzealous cache busting for e.g. multiple conda environments because the env prefix is included via `-I$PREFIX/include`, so multiple conda envs will not share a cache (maybe that's a good thing?). But other CFLAGS that change at the Python level may well be relevant.

### DIFF
--- a/ffcx/codegeneration/jit.py
+++ b/ffcx/codegeneration/jit.py
@@ -9,6 +9,7 @@ import io
 import logging
 import os
 import re
+import sysconfig
 import tempfile
 import time
 from contextlib import redirect_stdout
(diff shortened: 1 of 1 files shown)


garth-wells (approved):

garth-wells:
I think the cache key is missing the float type.

PR #532, opened 2022-09-01
Fix expression signature for expressions with ufl domain from constant

jorgensd:
Resolves Resolve https://github.com/FEniCS/dolfinx/issues/2342


nate-sime (approved):

chrisrichardson (approved):
Seems fine to me

PR #535, opened 2022-09-14
Bump CMake to 3.19

jorgensd:


jhale:
Can you explain the change from same minor version to any newer version?

cmake/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ install(TARGETS ${PROJECT_NAME}
 include(CMakePackageConfigHelpers)
 write_basic_package_version_file("${PROJECT_NAME}ConfigVersion.cmake"
                                  VERSION ${PROJECT_VERSION}
-                                 COMPATIBILITY SameMinorVersion)
+                                 COMPATIBILITY AnyNewerVersion)
```
garth-wells, line 26: This change doesn't look right. We don't guarantee backward compatibility for all future versions.

jorgensd:
> Can you explain the change from same minor version to any newer version?

From the documentation:

The COMPATIBILITY mode AnyNewerVersion means that the installed package version will be considered compatible if it is newer or exactly the same as the requested version. This mode should be used for packages which are fully backward compatible, also across major versions. If SameMajorVersion is used instead, then the behavior differs from AnyNewerVersion in that the major version number must be the same as requested, e.g. version 2.0 will not be considered compatible if 1.0 is requested. This mode should be used for packages which guarantee backward compatibility within the same major version. If SameMinorVersion is used, the behavior is the same as SameMajorVersion, but both major and minor version must be the same as requested, e.g version 0.2 will not be compatible if 0.1 is requested. If ExactVersion is used, then the package is only considered compatible if the requested version matches exactly its own version number (not considering the tweak version). For example, version 1.2.3 of a package is only considered compatible to requested version 1.2.3. This mode is for packages without compatibility guarantees. If your project has more elaborated version matching rules, you will need to write your own custom ConfigVersion.cmake file instead of using this macro.
New in version 3.19: The version file generated by AnyNewerVersion, SameMajorVersion and SameMinorVersion arguments of COMPATIBILITY handle the version range if any is specified (see [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package) command for the details). ExactVersion mode is incompatible with version ranges and will display an author warning if one is specified.

If we want to let Basix, FFCx, UFCx and DOLFINx to move independently, we should use AnyNewerVersion, as we could release a new version of Basix (0.6.0), that is compatible with DOLFINx (0.5.1), and with no changes in UFCx

cmake/CMakeLists.txt
jorgensd: Same minor doesn’t seem right either. The point is that  when finding a package should request either the exact version you need or the range of versions you support, which might span both minor or major versions

jorgensd: @garth-wells  from https://iamsorush.com/posts/cpp-cmake-config/
AnyNewerVersion means CMake imports the package if its version is newer or the same version as the requested one. For more details, see [CMake Manual](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file).

Which means it is up the package requesting UFCx, Basix or DOLFINx to request compatible version ranges

jhale:
Thanks for the explanation @jorgensd!

At the current rapid stage of development (pre 1.0.0) we are treating minor versions numbers as API breaking. Consequently, Basix 0.6.0 will not be compatible with DOLFINx 0.5.1.

So, I do think `SameMinorVersion` is correct.

When we get to `1.0.0` we will begin using the major number for API breaking changes, so we may need to change to `SameMajorVersion`?

We can chat on Slack when you're available, might be easier.

jorgensd:
> Thanks for the explanation @jorgensd!
> 
> At the current rapid stage of development (pre 1.0.0) we are treating minor versions numbers as API breaking. Consequently, Basix 0.6.0 will not be compatible with DOLFINx 0.5.1.
> 
> So, I do think `SameMinorVersion` is correct.
> 
> When we get to `1.0.0` we will begin using the major number for API breaking changes, so we may need to change to `SameMajorVersion`?
> 
> We can chat on Slack when you're available, might be easier.

Lets chat on monday.

jhale:
@jorgensd It would be nice to get this and the DOLFINx equivalent merged in as we move towards a 0.6.0 release.

cmake/CMakeLists.txt
jorgensd: We have decided that it is up to the developer of the external library to pin compatible version ranges.

.github/workflows/dolfin-tests.yml
```diff
@@ -10,7 +10,7 @@ on:
     inputs:
       dolfinx_ref:
         description: "DOLFINx branch or tag"
-        default: "main"
+        default: "dokken/cmake_version_range"
```
jorgensd, line 13:
```suggestion
        default: "main"
```

PR #539, opened 2022-10-07
Raise on invalid function name

michalhabera:


garth-wells:
@michalhabera any chance of some tests?

garth-wells:
@michalhabera ping again, could you add a test?

michalhabera:
> @michalhabera ping again, could you add a test?

What should the test do? The PR adds better explaining RuntimeError for invalid naming. It does not change
when the code fails, only how it fails.

jhale:
You can get pytest to test that an exception is raised or not? I can't think of anything else.

garth-wells:
@michalhabera check that exception is raised for an invalid name.

michalhabera:
Exceptions were caught and re-thrown in `jit.py`, so in order to preserve its type needs another try block. Moreover, Coefficients have user-given names only in dolfinx, so test at the FFCx level requires monkey patching, see the test.
Now you see why "testing" this is not very neat.

garth-wells (approved):

PR #566, opened 2023-04-11
Support facet integrals with vertex scheme

jorgensd:
I think we are getting to the point where this should really be in basix, as it would be easier to support more complicated cells, such as prisms, if we work directly with basix info.

Reported at: https://fenicsproject.discourse.group/t/set-integration-rule-to-closed-vertex/10695/5?u=dokken


garth-wells:
@jorgensd could a test be added?

jorgensd:
> @jorgensd could a test be added?

Test added

PR #571, opened 2023-06-07
Use `num_sub_entities(...)` (new function in UFL)

mscroggs:
instead of `ufl.cell.num_cell_entities[cell.cellname()][...]`.

This new function is added to UFL in https://github.com/FEniCS/ufl/pull/168


mscroggs:
One of the demos is failing (which is odd as it was passing with the UFL update before today.

I've set the demo to xfail and opened https://github.com/FEniCS/ffcx/issues/579 so that I can merge this, as FFCx is currently broken without this patch

PR #583, opened 2023-07-10
Fix bug that led to Quadrature Elements not being blocked.

jhale:
The order of the logic in element_interface.py led to the built in FFCx type QuadratureElement always being handled 
prior to the call to Basix's block element. This PR changes this to handle elements in order of higher-order types, FFCx types, and then finally UFL types.

It's not clear to be what the modern 'basix' way to handle QuadratureElements might be (regarding the warning).

### DIFF
--- a/test/test_blocked_elements.py
+++ b/test/test_blocked_elements.py
@@ -9,6 +9,7 @@ import numpy as np
 import ffcx
 import ffcx.codegeneration.jit
 import basix.ufl
+import ufl
 
 
 def test_finite_element(compile_args):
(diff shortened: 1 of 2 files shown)


test/test_blocked_elements.py
```diff
@@ -111,3 +112,38 @@ def test_tensor_element(compile_args):
         ufcx_dofmap.tabulate_entity_dofs(vals_ptr, 0, v)
         assert vals[0] == v
     assert ufcx_dofmap.num_sub_dofmaps == 4
+
+
+def test_vector_quadrature_element(compile_args):
+    ufl_element = ufl.VectorElement(ufl.FiniteElement("Quadrature", "tetrahedron", degree=2, quad_scheme="default"))
```
jhale, line 118: Can I do this without UFL?

jhale:
I've figured out how to construct the element with modern FFCx and Basix element constructors (rather than UFL) so I'll make that change first.

test/test_blocked_elements.py
jhale: Going to keep this as it is to test conversions work

Issue #447, opened 2022-01-25
Form with domain markers creates duplicated kernels


IgorBaratta:
MWE:
```python
coord_element = VectorElement("Lagrange", interval, 1)
mesh = Mesh(coord_element)

element = FiniteElement("Lagrange", interval, 1)
V = FunctionSpace(mesh, element)

u = TrialFunction(V)
v = TestFunction(V)

a = inner(u, v) * dx((1,2))
```

Produces 2 identitical kernels:

```c
// Code for integral integral_8799b8ef6f5269ca9a672cec402eefb7d5222475

void tabulate_tensor_integral_8799b8ef6f5269ca9a672cec402eefb7d5222475(double* restrict A,
                                    const double* restrict w,
                                    const double* restrict c,
                                    const double* restrict coordinate_dofs,
                                    const int* restrict entity_local_index,
                                    const uint8_t* restrict quadrature_permutation)
{
    // Quadrature rules
    static const double weights_4a8[2] = { 0.5, 0.5 };
    // Precomputed values of basis functions and precomputations
    // FE* dimensions: [permutation][entities][points][dofs]
    static const double FE0_C0_Q4a8[1][1][2][2] =
        { { { { 0.7886751345948129, 0.2113248654051871 },
              { 0.2113248654051871, 0.7886751345948129 } } } };
    static const double FE1_C0_D1_Q4a8[1][1][1][2] = { { { { -1.0, 1.0 } } } };
    // Quadrature loop independent computations for quadrature rule 4a8
    const double J_c0 = coordinate_dofs[0] * FE1_C0_D1_Q4a8[0][0][0][0] + coordinate_dofs[3] * FE1_C0_D1_Q4a8[0][0][0][1];
    double sp_4a8[1];
    sp_4a8[0] = fabs(J_c0);
    for (int iq = 0; iq < 2; ++iq)
    {
        const double fw0 = sp_4a8[0] * weights_4a8[iq];
        double t0[2];
        for (int i = 0; i < 2; ++i)
            t0[i] = fw0 * FE0_C0_Q4a8[0][0][iq][i];
        for (int i = 0; i < 2; ++i)
            for (int j = 0; j < 2; ++j)
                A[2 * i + j] += FE0_C0_Q4a8[0][0][iq][j] * t0[i];
    }
}


void tabulate_tensor_integral_b1ad34778d813d876edcdf491953bf64df0b3699(double* restrict A,
                                    const double* restrict w,
                                    const double* restrict c,
[...]

PR #589, opened 2023-08-11
Remove generated kernel duplication

jorgensd:
```python
from ufl import FiniteElement, TrialFunction, TestFunction, dx, Mesh, VectorElement, triangle, FunctionSpace, inner, grad
import ufl.algorithms
import ufl.classes
mesh = Mesh(VectorElement("Lagrange", triangle, 1))
V = FunctionSpace(mesh, FiniteElement("Lagrange", triangle, 1))
u = TrialFunction(V)
v = TestFunction(V)
t = tuple([i for i in range(1000)])
a = inner(u, v)*dx(t) + inner(grad(u), grad(v))*dx(2) + inner(u, v)*dx
```
generates a `c` file with 
```python
  // Quadrature rules
  static const double weights_48e[3] = { 0.1666666666666667, 0.1666666666666667, 0.1666666666666667 };
  // Precomputed values of basis functions and precomputations
  // FE* dimensions: [permutation][entities][points][dofs]
  static const double FE3_C0_Q48e[1][1][3][3] =
    { { { { 0.6666666666666667, 0.1666666666666666, 0.1666666666666667 },
          { 0.1666666666666667, 0.1666666666666666, 0.6666666666666665 },
          { 0.1666666666666668, 0.6666666666666665, 0.1666666666666667 } } } };
  static const double FE4_C0_D10_Q48e[1][1][1][3] = { { { { -1.0, 1.0, 0.0 } } } };
  static const double FE4_C1_D01_Q48e[1][1][1][3] = { { { { -1.0, 0.0, 1.0 } } } };
  // Quadrature loop independent computations for quadrature rule 48e
  double J_c0 = 0.0;
  double J_c3 = 0.0;
  double J_c1 = 0.0;
  double J_c2 = 0.0;
  for (int ic = 0; ic < 3; ++ic)
  {
    J_c0 += coordinate_dofs[ic * 3] * FE4_C0_D10_Q48e[0][0][0][ic];
    J_c3 += coordinate_dofs[ic * 3 + 1] * FE4_C1_D01_Q48e[0][0][0][ic];
    J_c1 += coordinate_dofs[ic * 3] * FE4_C1_D01_Q48e[0][0][0][ic];
    J_c2 += coordinate_dofs[ic * 3 + 1] * FE4_C0_D10_Q48e[0][0][0][ic];
  }
  double sp_48e[4];
  sp_48e[0] = J_c0 * J_c3;
  sp_48e[1] = J_c1 * J_c2;
  sp_48e[2] = sp_48e[0] + -1 * sp_48e[1];
  sp_48e[3] = fabs(sp_48e[2]);
  for (int iq = 0; iq < 3; ++iq)
  {
    const double fw0 = sp_48e[3] * weights_48e[iq];
    double t0[3];
    for (int i = 0; i < 3; ++i)
      t0[i] = fw0 * FE3_C0_Q48e[0][0][iq][i];
    for (int i = 0; i < 3; ++i)
      for (int j = 0; j < 3; ++j)
        A[3 * i + j] += FE3_C0_Q48e[0][0][iq][j] * t0[i];
  }
```
repeated 1000 times.
This makes the code generation slow and not scalable with an increasing number of subdomains.
Running `ffcx` with a 1000 subdomains takes about 3.8 seconds, while 10000 takes ~43 seconds.

This PR depends on: 
https://github.com/FEniCS/ufl/pull/92
and resolves
https://github.com/FEniCS/ffcx/issues/447

The current PR with 1000 subdomains runs in 0.9 seconds and 10000 subdomains runs in 8 seconds.
The `c`-file in this branch is 570 lines with 10000 subdomains, while it is over 600 000 lines on the main branch


jorgensd:
Note that there is now an implicit dependency in the generation of the offsets. The list in Line 102 of forms.py is dependent on the `dolfinx::fem::IntegralType`-ordering.
This means that if we change the `IntegralType` (either value for an entry or add more, the generation of offsets will be wrong).
This is exposed in: 
https://github.com/FEniCS/ffcx/pull/589/commits/2fda42c4cc276b39251fa52639cb3ae1b7c6678b
and was introduced in DOLFINx in
https://github.com/FEniCS/dolfinx/pull/2744/files

chrisrichardson:
I think the ordering is in the ufcx enum, not in dolfinx. I agree this is a weakness which should be addressed, though I am not quite sure how. One idea would be to provide a function (fixed code not generated). Is it worth it? I feel it is duplication - two ways of doing the same thing...

jorgensd:
> I think the ordering is in the ufcx enum, not in dolfinx. I agree this is a weakness which should be addressed,

As it is used in DOLFINx, 

> I think the ordering is in the ufcx enum, not in dolfinx. I agree this is a weakness which should be addressed, though I am not quite sure how. One idea would be to provide a function (fixed code not generated). Is it worth it? I feel it is duplication - two ways of doing the same thing...

I see, it uses the `ufcx.h` file to get `exterior_facet` etc in one place in `dolfinx/fem/utils.h` (all other places in the file uses `IntegralType::...`, so that threw me off.
I think the way to fix this is instead of looping over strings when building the offset, it should base this on the `ufch.x` header enum, or smth.

jorgensd:
@chrisrichardson could we go along at get this merged (and the PR in ufl), as it is getting quite tedious to keep up to date with rewrites of ffcx, and I haven't seen anyone having any objections regarding this PR

ffcx/codegeneration/form.py
```diff
@@ -90,8 +90,17 @@ def generator(ir, options):
     integral_offsets = [0]
     # Note: the order of this list is defined by the enum ufcx_integral_type in ufcx.h
     for itg_type in ("cell", "exterior_facet", "interior_facet"):
-        integrals += [L.AddressOf(L.Symbol(itg)) for itg in ir.integral_names[itg_type]]
-        integral_ids += ir.subdomain_ids[itg_type]
+        unsorted_integrals = []
```
chrisrichardson, line 93: Aren't they already sorted? Python `set()` (`representation.py` line 575)

jorgensd: No, they are not. Without this, `test_jit_forms.py::test_subdomains` failed.

ffcx/ir/representation.py
```diff
@@ -562,34 +561,22 @@ def _compute_form_ir(form_data, form_id, prefix, form_names, integral_names, ele
...
-                    assert isinstance(itg_data.subdomain_id, numbers.Integral)
-                    ir["subdomain_ids"][integral_type] += [itg_data.subdomain_id]
-                    ir["integral_names"][integral_type] += [integral_names[(form_id, itg_index)]]
+    ir["subdomain_ids"] = {itg_type: {} for itg_type in ufcx_integral_types}
+    ir["integral_names"] = {itg_type: {} for itg_type in ufcx_integral_types}
+    for itg_index, itg_data in enumerate(form_data.integral_data):
+        # UFL is using "otherwise" for default integrals (over whole mesh)
+        # but FFCx needs integers, so otherwise = -1
+        integral_type = itg_data.integral_type
+        subdomain_ids = set(sid if sid != "otherwise" else -1 for sid in itg_data.subdomain_id)
```
chrisrichardson, line 575: Is there a reason for using `set()` - can the subdomain_id be repeated?

test/test_jit_forms.py
```diff
@@ -879,3 +879,27 @@ def test_manifold_derivatives(compile_args):
...
+    `integrals_` and `integral_ids_`
+    """
+    mesh = ufl.Mesh(ufl.VectorElement("Lagrange", ufl.triangle, 1))
+    V = ufl.FunctionSpace(mesh, ufl.FiniteElement("Lagrange", ufl.triangle, 1))
+    u = ufl.TrialFunction(V)
+    v = ufl.TestFunction(V)
+    a = ufl.inner(u, v) * ufl.dx((1, 2, 3)) + ufl.inner(ufl.grad(u), ufl.grad(v)) * ufl.dx(2) + ufl.inner(u, v) * ufl.dx
+    compiled_forms, module, _ = ffcx.codegeneration.jit.compile_forms(
+        [a], cffi_extra_compile_args=compile_args)
+    # NOTE: This assumes that the first integral type is cell integrals, see UFCx.h
```
chrisrichardson, line 901: You can get the enum from `module.lib` to avoid an implicit assumption here.

ffcx/codegeneration/form.py
```diff
@@ -90,8 +90,17 @@ def generator(ir, options):
...
+        unsorted_ids = []
+        for key, name in ir.integral_names[itg_type].items():
+            for subdomain_id in ir.subdomain_ids[itg_type][key]:
+                unsorted_integrals += [L.AddressOf(L.Symbol(name))]
+                unsorted_ids += [subdomain_id]
+
+        id_sort = numpy.argsort(unsorted_ids)
+        integrals += [unsorted_integrals[id_sort[i]] for i in range(len(unsorted_integrals))]
+        integral_ids += [unsorted_ids[id_sort[i]] for i in range(len(unsorted_integrals))]
```
chrisrichardson, line 102:
```suggestion
        integral_ids += [unsorted_ids[i] for i in id_sort]
```

ffcx/ir/representation.py
```diff
@@ -562,34 +561,22 @@ def _compute_form_ir(form_data, form_id, prefix, form_names, integral_names, ele
...
-                    ir["subdomain_ids"][integral_type] = [-1] + ir["subdomain_ids"][integral_type]
-                    ir["integral_names"][integral_type] = [
-                        integral_names[(form_id, itg_index)]] + ir["integral_names"][integral_type]
-                elif itg_data.subdomain_id < 0:
-                    raise ValueError("Integral subdomain ID must be non-negative.")
-                else:
-                    assert isinstance(itg_data.subdomain_id, numbers.Integral)
-                    ir["subdomain_ids"][integral_type] += [itg_data.subdomain_id]
-                    ir["integral_names"][integral_type] += [integral_names[(form_id, itg_index)]]
+    ir["subdomain_ids"] = {itg_type: {} for itg_type in ufcx_integral_types}
```
chrisrichardson, line 569: Do they need to be `dicts` ? what about:
```suggestion
    ir["subdomain_ids"] = {itg_type: [] for itg_type in ufcx_integral_types}
```
since there is no need to store by ìtg_index` they can just be concatenated to a list.

chrisrichardson: Later on, then:
````
        ir["subdomain_ids"][integral_type] += list(subdomain_ids)
        for i in range(len(subdomain_ids)):
            ir["integral_names"][integral_type] += [integral_names[(form_id, itg_index)]]
````

File: ffcx/ir/representation.py (hunk 2 above)
Comment (jorgensd) [reply ffcx/ir/representation.py:575]: I think this was to encapsulate that subdomain_id can be both an integer or an arbitrary legnth tuple of integers. Do you have a better suggestion?

Comment (chrisrichardson) [reply ffcx/ir/representation.py:575]: But how does it work if `itg_data.subdomain_id` is an integer? The comprehension will fail...

Comment (chrisrichardson) [reply ffcx/ir/representation.py:575]:
```suggestion
        subdomain_ids = [sid if sid != "otherwise" else -1 for sid in itg_data.subdomain_id]
```

chrisrichardson (approved):
Don't forget CI changes before merging.

.github/workflows/dolfin-tests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Install UFL and Basix (default branches/tags)
         if: github.event_name != 'workflow_dispatch'
         run: |
-          python3 -m pip install git+https://github.com/FEniCS/ufl.git
+          python3 -m pip install git+https://github.com/FEniCS/ufl.git@dokken/group_integrals
```
mscroggs, line 49:
```suggestion
          python3 -m pip install git+https://github.com/FEniCS/ufl.git
```

.github/workflows/pythonapp.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: Install FEniCS dependencies (Python)
         run: |
-          python -m pip install git+https://github.com/FEniCS/ufl.git
+          python -m pip install git+https://github.com/FEniCS/ufl.git@dokken/group_integrals
```
mscroggs, line 51:
```suggestion
          python -m pip install git+https://github.com/FEniCS/ufl.git
```

mscroggs:
Now that https://github.com/FEniCS/ufl/pull/92 is merged, I think we need to merge this asap

PR #591, opened 2023-08-13
Remove switch/case from generated code

chrisrichardson:
This removes the function interfaces in ufcx for entity_dofs and form integrals, which are replaced with static data.

### DIFF
--- a/test/test_add_mode.py
+++ b/test/test_add_mode.py
@@ -13,17 +13,21 @@ import ufl
 from ffcx.naming import cdtype_to_numpy, scalar_to_value_type
 
 
-@pytest.mark.parametrize(
-    "mode", ["double", "float", "long double", "double _Complex", "float _Complex"]
-)
+@pytest.mark.parametrize("mode",
+                         [
+                             "double",
+                             "float",
+                             "long double",
+                             "double _Complex",
+                             "float _Complex"
+                         ])
 def test_additive_facet_integral(mode, compile_args):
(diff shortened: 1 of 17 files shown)


test/test_add_mode.py
```diff
@@ -13,33 +13,32 @@
...
-    assert ids[0] == -1
+    integral_offsets = form0.form_integral_offsets
+    ex = module.lib.exterior_facet
+    num_integrals = integral_offsets[ex + 1] - integral_offsets[ex]
+    assert num_integrals == 1
+    integral_id = form0.form_integral_ids[integral_offsets[ex]]
+    assert integral_id == -1
 
-    default_integral = form0.integrals(module.lib.exterior_facet)[0]
+    default_integral = form0.form_integrals[integral_offsets[ex]]
```
jorgensd, line 41: This here is implicitly tied to the ordering of the enum in DOLFINx, which I'm not sure is a good idea. See: https://github.com/FEniCS/ffcx/pull/589#issuecomment-1676768811

chrisrichardson: The order is explicitly defined in ufcx.h

jorgensd: It is still an issue for whenever we generate the offsets (as that is based on strings in a for-loop, not the enum).

chrisrichardson: I have annotated the point where this happens in ffcx.

Issue #590, opened 2023-08-11
Move implementation of `QuadratureElement` to `basix.ufl`


mscroggs:
It's currently confusing that most elements can be imported from `basix.ufl`, but this one must be imported from `ffcx.element_interface`

PR #592, opened 2023-08-14
Move QuadratureElement and RealElement implementations to Basix

mscroggs:
Resolves https://github.com/FEniCS/ffcx/issues/590.

Can be merged after https://github.com/FEniCS/basix/pull/682.


ffcx/element_interface.py
```diff
@@ -99,334 +59,23 @@ def map_facet_points(points: npt.NDArray[np.float64], facet: int, cellname: str)
...
-        """True if the discontinuous version of the element is used."""
-        return False
+# TODO: remove this deprecated function
+def QuadratureElement(
+    cellname: str, value_shape: typing.Tuple[int, ...], scheme: typing.Optional[str] = None,
+    degree: typing.Optional[int] = None, points: typing.Optional[npt.NDArray[np.float64]] = None,
+    weights: typing.Optional[npt.NDArray[np.float64]] = None, mapname: str = "identity"
+) -> basix.ufl._ElementBase:
+    warnings.warn(
+        "ffcx.element_interface.QuadratureElement is deprecated and will be removed after December 2023. "
```
mscroggs, line 69: If we agree with the date for removal give here, we should open an issue to remind us to actually delete these functions

ffcx/element_interface.py
```diff
@@ -24,45 +20,9 @@ def convert_element(element: ufl.finiteelement.FiniteElementBase) -> basix.ufl._
...
-    elif hasattr(ufl, "MixedElement") and isinstance(element, ufl.MixedElement):
-        return basix.ufl.mixed_element([_cached_conversion(e) for e in element.sub_elements()])
-    elif hasattr(ufl, "EnrichedElement") and isinstance(element, ufl.EnrichedElement):
-        return basix.ufl.enriched_element([_cached_conversion(e) for e in element._elements])
-    elif element.family() == "Quadrature":
-        return QuadratureElement(element.cell().cellname(), element.value_shape(), scheme=element.quadrature_scheme(),
-                                 degree=element.degree())
-    elif element.family() == "Real":
-        return RealElement(element)
-    else:
```
jorgensd, line 65: So the reason this is not needed anymore is due to everything being in basix (where all caching now happens)?
If so, this is rather neat

mscroggs: Yes. One of the main reasons for this PR was this this code is duplicated from code that already exists in Basix, except with the addition of QuadratureElement and RealElement.

mscroggs: Except Basix doesn't cache. Creating elements in UFL (instead of directly with basix.ufl) is deprecated. Leaving in part of the function so that we could cache the result of a deprecated function seemed unnecessary as it'll be deleted soon anyway

jorgensd: It does? https://github.com/FEniCS/basix/blob/7daa8bf1eb8774e685a6e30a91e585e409f03d87/python/basix/ufl.py#L1585

mscroggs: Sort of. It caches pure Basix elements, but not Mixed, Quadrature, etc. Although the pure Basix elements will be the most costly to recreate so I guess this is good enough

ffcx/element_interface.py
mscroggs: Done: https://github.com/FEniCS/ffcx/issues/605

PR #594, opened 2023-08-25
Rename CNodes as LNodes, add typing, process AST separately from creating it.

chrisrichardson:
This PR adresses removes CNodes, and replaces with a much slimmed-down "LNodes" which contains the basics:

* Variables and Arrays of `Scalar` (same as tensor), `Real` (same as geometry) and `Int`
* Basic math operations +-*/
* Math functions, including conditional statements
* ForRange loops

but removes almost everything else.

* All `Symbols` must now have a `DataType`

Additionally, the Abstract Syntax Tree (AST) which is generated is now processed separately in `c_implementation.py`, which will allow us to post-process into other languages/libraries in the future.

This is quite a large PR, unfortunately, but I think it will move things in the right direction.


chrisrichardson:
I just added some formatting for complex numbes and variable precision. Standard C needs the `1.0+1.0*I` rather than `1.0+1.0j` although modern compilers seem to accept either.

garth-wells (approved):

PR #600, opened 2023-09-05
Fix issue #594

chrisrichardson:
Add parentheses to negative expressions when needed.


francesco-ballarin:
Thanks @chrisrichardson!

PR #604, opened 2023-09-06
Fix #602

### LINKED ISSUE #602: warning: implicit declaration of function ‘real’; did you mean ‘creal’?
Generated code for a form is giving this warning. See eg https://github.com/FEniCS/basix/actions/runs/6094169143/job/16535151559?pr=682

mscroggs:
After this PR:

- If `real` or `conj` is called on a real number, nothing is done
- If `imag` is called on a real number, 0 is returned

This is a slightly hacky fix for #602.

Happy for this to be closed if a better fix exists

edit: this feels even hackier now that it's been edited to make mypy pass

Close #602. Close https://github.com/FEniCS/dolfinx/issues/2757

### DIFF
--- a/ffcx/codegeneration/C/c_implementation.py
+++ b/ffcx/codegeneration/C/c_implementation.py
@@ -313,7 +313,7 @@ class CFormatter(object):
             func = func.split(" ", 1)[1]
             if func == "()":
                 assert len(c.args) == 1
-                return self.c_format(c.args[0])
+                return "(" + self.c_format(c.args[0]) + ")"
             elif func == "0":
                 return "0"
             else:
--- a/ffcx/codegeneration/C/c_implementation.py
+++ b/ffcx/codegeneration/C/c_implementation.py
@@ -34,6 +34,8 @@ math_table = {
         "max_value": "fmax",
         "bessel_y": "yn",
         "bessel_j": "jn",
+        "real": None,
+        "conj": None,
     },
     "float": {
         "sqrt": "sqrtf",
@@ -59,6 +61,8 @@ math_table = {
         "max_value": "fmaxf",
         "bessel_y": "yn",
         "bessel_j": "jn",
+        "real": None,
+        "conj": None,
     },
     "long double": {
         "sqrt": "sqrtl",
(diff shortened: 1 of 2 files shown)


garth-wells:
What is `real` is the generated code supposed to do? Is it an artefact, or should it be `creal`?

chrisrichardson:
It should not be `creal` as the input is `double` not `_Complex double`. It should not be anything. But ufl is asking for the real part of a number that is already real.

mscroggs:
> What is `real` is the generated code supposed to do? Is it an artefact, or should it be `creal`?

`real` is replaced with `creal` for double complex and `crealf` for float complex. It's currently not replaced for real types. This PR makes a replacement happen for these types

chrisrichardson:
Maybe use "" instead of None, then it will just use parentheses, which have no effect. Also a bit hacky.

mscroggs:
> Maybe use "" instead of None, then it will just use parentheses, which have no effect. Also a bit hacky.

mypy also didn't like None. I went with `"SPECIAL_CASE ()"` and `"SPECIAL_CASE 0` (as the latter will also allow dealing with imag(real type) = 0. But open to better suggestions

chrisrichardson:
How about picking it up a bit sooner (before consulting the math table):
```
if c.args[0].dtype == L.DataType.REAL:
    if c.function in ("real", "conj"):
        return "(" + self.c_format(c.args[0]) + ")"
    elif c.function == "imag":
        return "0"
```

chrisrichardson:
Or maybe we should handle it in `lnodes.py` before it gets here...
```
def _math_function(op, *args):
    name = op._ufl_handler_name_
    dtype = args[0].dtype
    if name in ("conj", "real") and dtype == DataType.REAL:
        assert len(args) == 1
        return args[0]
    if name == "imag" and dtype == DataType.REAL:
        assert len(args) == 1
        return LiteralFloat(0.0)
    return MathFunction(op._ufl_handler_name_, args)

```

chrisrichardson:
I think this is the least "hacky", let's hope it passes all the tests...

chrisrichardson (approved):

PR #609, opened 2023-09-08
Remove precision option

chrisrichardson:
Removes precision option and just uses `str()` in Python to print to the internal precision.


garth-wells (changes requested):
This is very problematic. We are ceding all control of precision, with no guarantee that `str` gives enough digits.

garth-wells (approved):
I think this is fine to get the DOLFINx CI (hopefully) green again.

We can then discuss refinements and alternatives more leisurely.

PR #629, opened 2023-10-31
Allow selection of spack tag

chrisrichardson:
Adds parameter to CI, so you can choose which spack tag to build, 0.6.0, 0.7.0, main etc.


.github/workflows/spack.yml
```diff
@@ -18,6 +18,10 @@ on:
         description: "Spack repository branch/tag to test"
         default: "develop"
         type: string
+      spack_tag:
```
garth-wells, line 21: This should be `ffcx_version` rather than `spack_tag`.

PR #642, opened 2023-12-19
[Sum Factorization - 1] Transition to 1D basis evaluation matrices for hexes and quads

IgorBaratta:
This PR aims to transition from 3D/2D to 1D basis evaluation matrices for hexahedral and quadrilateral elements.

To experiment with this new feature, users can activate it by setting `options["sum_factorization"] = True` or using the command line interface. It's important to note, however, that this modification does not fully implement sum factorization nor does it offer immediate performance improvements. These enhancements are part of our roadmap and will be addressed in an upcoming PR.

Some current limitations:
- FFCx does not automatically verify if the provided dofmap is in tensor product order. For an example of this, refer to the 'mass_action.py' demo.
- At present, only Lagrange elements are equipped with tensor factors.

A significant portion of this work was carried out in collaboration with @mscroggs.


ffcx/codegeneration/definitions.py
```diff
@@ -7,12 +7,51 @@
...
 
+from typing import List
 import ufl
 import ffcx.codegeneration.lnodes as L
 
 logger = logging.getLogger("ffcx")
 
 
+def create_nested_for_loops(indices: List[L.MultiIndex], body):
+    ranges = [r for idx in indices for r in idx.sizes]
```
chrisrichardson, line 18: Add a docstring

ffcx/codegeneration/definitions.py
```diff
@@ -79,8 +131,8 @@ def coefficient(self, t, mt, tabledata, quadrature_rule, access):
         assert begin < end
 
         # Get access to element table
-        FE = self.symbols.element_table(tabledata, self.entitytype, mt.restriction)
-        ic = self.symbols.coefficient_dof_sum_index
+        # FE = self.symbols.element_table(tabledata, self.entitytype, mt.restriction)
```
chrisrichardson, line 134: remove commented line

ffcx/codegeneration/definitions.py
```diff
@@ -79,8 +131,8 @@ def coefficient(self, t, mt, tabledata, quadrature_rule, access):
...
+        FE = self.symbols.table_access(tabledata, self.entitytype, mt.restriction, iq, ic)
```
chrisrichardson, line 135: Probably should be in `access.py` not in `symbols.py` ?

ffcx/codegeneration/integral_generator.py
```diff
@@ -446,17 +446,18 @@ def generate_block_parts(self, quadrature_rule: QuadratureRule, blockmap: Tuple,
...
-        for i in range(block_rank):
-            B_indices.append(arg_indices[i])
-        B_indices = list(B_indices)
+        iq_symbol = self.backend.symbols.quadrature_loop_index
+        iq = create_quadrature_index(quadrature_rule, iq_symbol)
 
         for blockdata in blocklist:
+
+            B_indices = []
+            args = [L.Symbol(name, L.DataType.INT) for name in ["i", "j", "k", "l"]]
```
chrisrichardson, line 455: This already exists in `symbols.py` as `argument_loop_index()` - should we use (or delete from `symbols.py`)?

ffcx/codegeneration/lnodes.py
```diff
@@ -327,18 +327,20 @@ def __repr__(self):
...
     def __init__(self, symbols: list, sizes: list):
         self.dtype = DataType.INT
         self.sizes = sizes
         self.symbols = [as_lexpr(sym) for sym in symbols]
         for sym in self.symbols:
             assert sym.dtype == DataType.INT
 
-        dim = len(sizes)
-        if dim == 0:
+        self.dim = len(sizes)
```
chrisrichardson, line 339: I'd rather not store `self.dim` since it is already the same as `len(self.sizes)`. We could add a method or `@property` to return it:
```
@property
def dim(self):
    return len(self.sizes)
```

ffcx/ir/representationutils.py
```diff
@@ -44,23 +46,54 @@ def id(self):
         return self.hash_obj.hexdigest()[-3:]
 
 
-def create_quadrature_points_and_weights(integral_type, cell, degree, rule, elements):
+def product(values):
```
chrisrichardson, line 49: Isn't there a `numpy` function to do this?

ffcx/ir/representationutils.py
```diff
@@ -44,23 +46,54 @@ def id(self):
...
+            elif cell.cellname() == "hexahedron":
+                tensor_factors = [
+                    create_quadrature("interval", degree, rule, elements)
+                    for _ in range(3)]
+            pts = np.array([
+                tuple(i[0] for i in p)
+                for p in itertools.product(*[f[0] for f in tensor_factors])
+            ])
+            wts = np.array([
+                product(i for i in p)
```
chrisrichardson, line 78:
```suggestion
                product(p)
```

ffcx/ir/representationutils.py
chrisrichardson: Why is it `- i`? (can multiply by `-1**n` if needed).

ffcx/codegeneration/definitions.py
IgorBaratta: Access is not part of `FFCXBackendDefinitions`.
I think it would require some restructuring.

chrisrichardson (approved):

PR #647, opened 2024-01-08
Use NumPy strings for scalar type

### LINKED ISSUE #595: Use numpy-like descriptions for scalar type in command line
Instead of using C types, it would be neater to use numpy types for the scalar type, when needed.
e.g.
`ffcx --scalar_type=complex128 Form.py` rather than `ffcx --scalar_type="double _Complex" Form.py`

The translation to implementation scalar type would take place in the "C" language implementation part.

garth-wells:
Use `float32`, `float64`, `complex64`, `complex128`, etc, for scalar type.

This is simpler than pass the C type, e.g. `float _Complex` (easier to pass options without spaces) and improves readability. Will also support non-C backends.  

Resolves #595.


chrisrichardson:
See Issue #595

chrisrichardson (approved):

PR #651, opened 2024-01-16
Avoid nesting conditions and statements with many operands

IgorBaratta:
This PR fixes issue [#2991](https://github.com/FEniCS/ffcx/issues/652).

Also it simplifies how we create statements with LNodes in the integral generator.
Before this PR, each statement could be nested and could be understood as a tree of statements:
```
A = OP1(OP2(X, OP3(Y, Z)), W)
```
With this PR we create statements with 2 operands and 1 destination:
```
A0 = OP3(Y, Z)
A1 = OP2(X, A0)
A2 = OP1(A1, W)
```

### DIFF
--- a/ffcx/codegeneration/C/c_implementation.py
+++ b/ffcx/codegeneration/C/c_implementation.py
@@ -156,6 +156,8 @@ class CFormatter(object):
             return dtype_to_c_type(self.real_type)
         if dtype == L.DataType.INT:
             return "int"
+        if dtype == L.DataType.BOOL:
+            return "bool"
         raise ValueError(f"Invalid dtype: {dtype}")
 
     def _format_number(self, x):
--- a/ffcx/codegeneration/expression_generator.py
+++ b/ffcx/codegeneration/expression_generator.py
@@ -333,39 +333,16 @@ class ExpressionGenerator:
                 # Get previously visited operands
                 vops = [self.get_var(op) for op in v.ufl_operands]
 
-                # get parent operand
-                pid = F.in_edges[i][0] if F.in_edges[i] else -1
-                if pid and pid > i:
-                    parent_exp = F.nodes.get(pid)['expression']
-                else:
-                    parent_exp = None
-
                 # Mapping UFL operator to target language
                 self._ufl_names.add(v._ufl_handler_name_)
                 vexpr = L.ufl_to_lnodes(v, *vops)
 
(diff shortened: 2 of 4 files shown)


ffcx/codegeneration/C/c_implementation.py
```diff
@@ -156,6 +156,8 @@ def _dtype_to_name(self, dtype) -> str:
             return dtype_to_c_type(self.real_type)
         if dtype == L.DataType.INT:
             return "int"
+        if dtype == L.DataType.BOOL:
+            return "bool"
```
chrisrichardson, line 160: I am not sure about using `bool` in C? Is it defined? Maybe in C99, but I think then you are supposed to include `stdbool.h`?

michalhabera: https://github.com/FEniCS/ffcx/blob/main/ffcx/codegeneration/ufcx.h#L26

michalhabera:
So this change disables the intermediate array storage, instead, all intermediates go to new local variable. This is certainly better for varying data types of intermediate symbols. I think the old array like storage was probably motivated by performance, but I cannot see why this argument should hold. Did you benchmark it?

IgorBaratta:
> So this change disables the intermediate array storage, instead, all intermediates go to new local variable. This is certainly better for varying data types of intermediate symbols. I think the old array like storage was probably motivated by performance, but I cannot see why this argument should hold. Did you benchmark it?

Yes, for most forms the C compiler generates the same code for the array like storage and scalar variables. In other cases the compiler can vectorize the code with local variables (SLP vectorization) but not with the array.

michalhabera:
> > So this change disables the intermediate array storage, instead, all intermediates go to new local variable. This is certainly better for varying data types of intermediate symbols. I think the old array like storage was probably motivated by performance, but I cannot see why this argument should hold. Did you benchmark it?
> 
> Yes, for most forms the C compiler generates the same code for the array like storage and scalar variables. In other cases the compiler can vectorize the code with local variables (SLP vectorization) but not with the array.

Right, SLP vectorizer probably does not work with arrays, that's is a good point. I've never seen SLP do anything useful in our codes in the past, but that was most likely due to the opaque array. I think this is a bugfix with a change in the right direction ;)

chrisrichardson (approved):
Looks good, as long as it passes the tests...

PR #653, opened 2024-01-17
Remove information about physical cells from elements

mscroggs:
See https://github.com/FEniCS/ufl/pull/249

Coupled with FEniCS/ufl#249, FEniCS/basix#772, FEniCS/dolfinx#2996.


demo/ProjectionManifold.py
```diff
@@ -22,10 +22,10 @@
                  inner)
 
 # Define element over this domain
-V = basix.ufl.element("RT", "triangle", 1, gdim=3)
-Q = basix.ufl.element("DG", "triangle", 0, gdim=3)
+V = basix.ufl.element("RT", "triangle", 1)
+Q = basix.ufl.element("DG", "triangle", 0)
 element = basix.ufl.mixed_element([V, Q])
-domain = Mesh(basix.ufl.element("Lagrange", "triangle", 1, shape=(3, ), gdim=3))
+domain = Mesh(basix.ufl.element("Lagrange", "triangle", 1, shape=(3, )))
```
jorgensd, line 28: This is an example where either the mesh, or the element has to know its block size (as the geometrical dimension of the mesh is interpreted from the shape).

mscroggs: Currently, the shape of a coordinate element is always (gdim, ) and I don't think it makes sense for it ever be different, so inferring gdim from shape seems sensible

jorgensd: I am thinking more towards the future, if we want to move block size out of the element and into the function space.
Then the coordinate element is no longer a useful abstraction, as it does not have sufficient information.

PR #657, opened 2024-01-22
Test demos with different scalar types

IgorBaratta:
Currently, demos are only tested in real mode.
Some issues are only picked up in complex mode when running the integration tests with dolfinx.

Only updated tests that required minimal effort to work with complex numbers.

### DIFF
--- a/demo/ReactionDiffusion.py
+++ b/demo/ReactionDiffusion.py
@@ -19,7 +19,7 @@
 # reaction-diffusion equation using simplified tuple notation.
 import basix.ufl
 from ufl import (Coefficient, FunctionSpace, Mesh, TestFunction, TrialFunction,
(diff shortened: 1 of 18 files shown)


demo/ReactionDiffusion.py
```diff
@@ -19,7 +19,7 @@
 # reaction-diffusion equation using simplified tuple notation.
 import basix.ufl
 from ufl import (Coefficient, FunctionSpace, Mesh, TestFunction, TrialFunction,
-                 dx, grad, inner)
+                 conj, dx, grad, inner)
```
chrisrichardson, line 22:
```suggestion
                 dx, grad, inner)
```

chrisrichardson (approved):

PR #663, opened 2024-02-01
Add Python 3.12 coverage to CI

garth-wells:
This showed up that we need to install `setuptools` for `cffi` JIT with Python 3.12.


jorgensd:
Tangential question: should we make import sorting blocking on CI?

PR #668, opened 2024-02-06
null pointer for kernels that are not defined

### LINKED ISSUE #666: Set undefined tabulate kernels to `NULL` in UFCx code

mscroggs:
Resolves #666


ffcx/codegeneration/C/integrals.py
```diff
@@ -56,6 +56,14 @@ def generator(ir, options):
     code["additional_includes_set"] = set()  # FIXME: Get this out of code[]
     code["tabulate_tensor"] = body
 
+    code["tabulate_tensor_float32"] = "NULL"
+    code["tabulate_tensor_float64"] = "NULL"
+    code["tabulate_tensor_longdouble"] = "NULL"
```
garth-wells, line 61: We should remove all the `long double' and stick with properly defined types. Ill-defined, platform-dependent types cause nothing but trouble.

PR #674, opened 2024-03-13
Fix a bug in Section equality check.

jorgensd:
Only add definitions of variable if unique.
Resolves https://github.com/FEniCS/dolfinx/issues/3102

### DIFF
--- a/test/test_submesh.py
+++ b/test/test_submesh.py
@@ -58,7 +58,7 @@ def compute_tensor(forms: list[ufl.form.Form], dtype: str, compile_args: list[st
         "complex128",
     ],
 )
-def test_submesh_codim0(dtype, compile_args):
+def test_multiple_mesh_codim0(dtype, compile_args):
     # Define coordinate element and element used in parent and sub-mesh
     element = basix.ufl.element("Lagrange", "triangle", 1)
     coordinate_element = basix.ufl.element("Lagrange", "triangle", 1, shape=(2,))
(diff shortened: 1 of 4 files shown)


IgorBaratta:
LGTM!
I'll try to cook up a test as discussed in slack.

IgorBaratta (approved):

test/test_submesh.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+@pytest.mark.parametrize(
+    "dtype",
+    [
+        "float64",
+        "complex128",
+    ],
+)
+def test_submesh_codim0(dtype, compile_args):
```
jpdean, line 61: Am I right in thinking that this issue is caused when two different `ufl.Mesh`es are used in a form that have the same coordinate element? If so, maybe this test and file should be renamed to better reflect this? To me it seems like it's testing we don't get code duplication, rather than testing submeshes

jpdean (approved):
Looking good to me

test/test_submesh.py
jorgensd: The only case when you will have two `ufl.Mesh` in the same form is whenever we are using sub-meshes.

jpdean: That's probably true, but if the user happened to have done something like

```
V = FunctionSpace(ufl.Mesh(coordinate_element), V_element)
W = FunctionSpace(ufl.Mesh(coordinate_element), W_element)
``` 

then we would still wan't that to work, right?

jorgensd, line 61:
```suggestion
def test_multiple_mesh_codim0(dtype, compile_args):
```

PR #675, opened 2024-03-28
Add support for mixed-dimensional kernels (codimension 1)

jpdean:
This PR adds support for mixed-dimensional kernels of codimension 1. It requires [this](https://github.com/FEniCS/ufl/pull/265) PR to work.

It would be nice to avoid passing `cell` to `analyse_modified_terminal`. Please let me know if anyone has any thoughts on this.


ffcx/codegeneration/definitions.py
```diff
@@ -190,7 +192,7 @@ def _define_coordinate_dofs_lincomb(
         iq_symbol = self.symbols.quadrature_loop_index
         ic = create_dof_index(tabledata, ic_symbol)
         iq = create_quadrature_index(quadrature_rule, iq_symbol)
-        FE, tables = self.access.table_access(tabledata, self.entitytype, mt.restriction, iq, ic)
+        FE, tables = self.access.table_access(tabledata, self.entitytype, mt.restriction, iq, ic, 0)
```
jorgensd, line 195: Maybe we should add a note here stating that due to the way we define the mixed dimensional code, i.e. that the integration domain for codim=1 has to be the "parent mesh", the coordinate element is  always of codim 0.

ffcx/ir/analysis/modified_terminals.py
```diff
@@ -187,6 +192,10 @@ def analyse_modified_terminal(expr):
     reference_value = None
     restriction = None
     averaged = None
+    if cell is not None and expr.ufl_domain() is not None:
+        codim = cell.topological_dimension() - expr.ufl_domain().topological_dimension()
```
jorgensd, line 196:
```suggestion
        codim = cell.topological_dimension() - expr.ufl_domain().topological_dimension()
        assert codim >= 0
```

ffcx/ir/analysis/modified_terminals.py
```diff
@@ -168,7 +173,7 @@ def strip_modified_terminal(v):
     return v
 
 
-def analyse_modified_terminal(expr):
+def analyse_modified_terminal(expr, cell=None):
```
jorgensd, line 176: Maybe we should call this `coordinate_cell`? I.e. the cell that the `coordinate-element` is using? 
I guess the doc-string should reflect this if we decide not to change the name.

jorgensd: Since we are at it, I think we should add type-hints here, i.e.

```python

from ufl.classes import (
...
    Cell,
...
)
from ufl.core.expr import Expr

def analyse_modified_terminal(expr: Expr, cell: typing.Optional[Cell]=None):
    ....
```

ffcx/ir/elementtables.py
```diff
@@ -341,7 +355,7 @@ def build_optimized_tables(
         # the dofmap offset may differ due to restriction.
 
         tdim = cell.topological_dimension()
```
jorgensd, line 357: Could we add a comment here as to what happens below? I guess what is going on is that if the integral is an interior facet or mixed dimensional integral, we need to compute the "facet-permutations", to ensure that the codim-1 cell is aligned with the codim-0 facet?

ffcx/ir/analysis/modified_terminals.py
cdaversin: I agree with Jørgen's comment (we need to make sure that `codim >= 0`), but maybe even better to have an explicit error message when it is not the case

ffcx/ir/integral.py
```diff
@@ -80,11 +80,17 @@ def compute_integral_ir(cell, integral_type, entitytype, integrands, argument_sh
...
             for i, v in S.nodes.items()
             if is_modified_terminal(v["expression"])
         }
 
+        # Check if we have a mixed-dimensional integral
+        is_mixed_dim = False
+        for domain in integrand.ufl_domains():
+            if domain.topological_dimension() != cell.topological_dimension():
+                is_mixed_dim = True
+
```
cdaversin, line 93: Is there a reason to use a boolean `(is_mixed_dim)` instead of computing `codim = cell.topological_dimension() - domain.topological_dimension()` ?

test/test_jit_forms.py
```diff
@@ -1074,3 +1074,73 @@ def test_integral_grouping(compile_args):
...
+
+    u = ufl.TrialFunction(V)
+    vbar = ufl.TestFunction(Vbar)
+
+    f = ufl.Coefficient(V)
+    g = ufl.Coefficient(Vbar)
+
+    ds = ufl.Measure("ds", domain=V_domain)
+
+    forms = [ufl.inner(f * g * u, vbar) * ds]
```
cdaversin, line 1107: I think it would be good to test expressions with gradients on coefficients from either mesh, to ensure that FacetJacobians, FacetGradients etc are computed correctly (jorgensd agrees)

michalhabera:
I think you've correctly noted that passing cell into `analyse_modified_terminal` is weird. It feels odd to store codim (which is the property of the element) to modified terminal, which has the purpose of defining the context in which the element is tabulated. In many places you already have the `cell` and the `element`, from which you can deduce the codimension, no?

For the analysis, I'd try to deduce the codim from cell and element and for the access you might need to store the codim as property of tabledata. Does that make sense or I am missing something?

jpdean:
Thanks for the feedback everyone! [This](https://github.com/FEniCS/dolfinx/pull/3209) PR means that permutations are now only needed for the codim 0 cell, which simplifies this PR quite a lot. I think this will address most of the comments here, but I'll work on fixing the others soon.

ffcx/ir/integral.py
jpdean: `is_mixed_dim` indicates if an integral contains any mixed-dimensional terms. In `build_optimized_tables`, we need to differentiate between the codim 0 element and the codim 1 element in a mixed-dimensional integral so that only the quadrature rule for the codim 0 element is permuted (see [here](https://github.com/FEniCS/ffcx/pull/675/files#diff-b2e2e37c190c9887e10625dad288f14dfc31c596ffab0ca77602d66399c0a019R367)). I've added a check that `codim > 0` when the codimension of the modified terminal is computed [here](https://github.com/FEniCS/ffcx/pull/675/files#diff-b2e2e37c190c9887e10625dad288f14dfc31c596ffab0ca77602d66399c0a019R358). An alternative approach might be to replace `is_mixed_dim` with something like `integral_codim` and rename `codim` to `element_codim`. Let me know which you prefer

test/test_jit_forms.py
jpdean: The test now uses `grad(u) * n`

test/test_jit_forms.py
```diff
@@ -1074,3 +1074,76 @@ def test_integral_grouping(compile_args):
...
+    kernel(
+        ffi.cast(f"{c_type}  *", A.ctypes.data),
+        ffi.cast(f"{c_type}  *", w.ctypes.data),
+        ffi.cast(f"{c_type}  *", c.ctypes.data),
+        ffi.cast(f"{c_xtype} *", coords.ctypes.data),
+        ffi.cast("int *", facet.ctypes.data),
+        ffi.cast("uint8_t *", perm.ctypes.data),
+    )
+
+    expected_result = np.array(
```
jorgensd, line 1138: It would be good to have some explanation as to what we are integrating here, and why this is the correct result.
I guess what you have is the facet normal on facet 0 (which is the one spanning (1.0,0) to (0.0, 1), i.e. a normal of (0.5, 0.5) multiplied by the gradient of the trial function on the parent cell.  It seems like both f and g are one (when looking at your coefficients, and it is not clear what order they are packed in). Could we use non-zero data here? 
This integral seems to simplify to `int_facet grad(u)[0]v + grad(u)[1]v * ds `

jpdean:
Are there any further issues anyone would like me to address?

test/test_jit_forms.py
```diff
@@ -1124,3 +1124,102 @@ def test_integral_grouping(compile_args):
         ]
     )
     assert len(unique_integrals) == 2
+
+
+# TODO Test 3D and non-simplex
```
jorgensd, line 1129: Remove TODO or add appropriate test.

jorgensd (approved):
LGTM! Only a small note on the test comment FIXME

garth-wells (approved):

PR #678, opened 2024-04-15
Adding more type hints and remove options when not used

jorgensd:
Cleaner interface with type hints, and only sending in all options when required

### DIFF
--- a/ffcx/compiler.py
+++ b/ffcx/compiler.py
@@ -91,12 +91,12 @@ def compile_ufl_objects(
     """Generate UFC code for a given UFL objects.
 
     Args:
-       ufl_objects: Objects to be compiled. Accepts elements, forms,
+        ufl_objects: Objects to be compiled. Accepts elements, forms,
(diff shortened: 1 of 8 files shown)


ffcx/compiler.py
```diff
@@ -81,29 +83,32 @@ def _print_timing(stage: int, timing: float):
...
-        object_names: TODO.
-        prefix: Prefix
-        options: Options
-        visualise: Toggle visualisation
+       ufl_objects: Objects to be compiled. Accepts elements, forms,
+          integrals or coordinate mappings.
+       object_names: Map from object Python id to object name
+       prefix: Prefix
+       options: Options
+       visualise: Toggle visualisation
```
jorgensd, line 99:
```suggestion
        ufl_objects: Objects to be compiled. Accepts elements, forms,
          integrals or coordinate mappings.
        object_names: Map from object Python id to object name
        prefix: Prefix
        options: Options
        visualise: Toggle visualisation
```

PR #679, opened 2024-04-15
Remove unused/unnecessary items from `IntegralIR`

jorgensd:
Removing the usage of `ir` as temporary storage of variables.


jhale:
Looks OK to me, please merge when ready.

jorgensd:
> Looks OK to me, please merge when ready.

Im happy with it. Just want to check with @mscroggs that he agrees with what Im removing (even if it is clearly unused, as seem by the tests).

PR #688, opened 2024-04-29
Write the Basix hash into generated code (or 0 if the element is not plain C++ Basix element)

mscroggs:
Required for https://github.com/FEniCS/dolfinx/issues/3169.

DOLFINx PR that follows on from this: https://github.com/FEniCS/dolfinx/pull/3186


ffcx/codegeneration/ufcx.h
```diff
@@ -162,7 +162,7 @@ extern "C"
     bool needs_facet_permutations;
 
     /// Get the coordinate element associated with the geometry of the mesh.
-    long int coordinate_element;
+    unsigned long int coordinate_element;
```
mscroggs, line 165: I'm getting a lot of `warning: integer constant is so large that it is unsigned` when compiling the generated code for this, but adjusting the type doesn't seem to make any difference

mscroggs: Adding `u` to the end of each literal int removed this warning

garth-wells (commented):
Could we add `_hash` to the variable names in `ufcx.h`?

garth-wells (approved):
Is `u` the right suffix? 

Looks like it's all rather messy in C, https://en.cppreference.com/w/c/language/integer_constant.

PR #689, opened 2024-04-30
Add Windows CI

jhale:
- Brings back macOS test.
- Makes pygraphviz truly optional.
- All C builds done in C17 standard mode.
- Support for Visual C compiler on Windows (does not support _Complex).
- Does not preclude use of ffcx on Windows for complex code generation with another compiler invoked via e.g. CMake or at the command line.


jhale:
@minrk Could you try this one Conda? The thing 'difficult' here is the use of CFFI.

ffcx/codegeneration/jit.py
```diff
@@ -118,11 +130,18 @@ def _compilation_signature(cffi_extra_compile_args=None, cffi_debug=None):
...
+            return str(s)
+
+    # TODO: SOABI not defined on win32
     return (
-        str(cffi_extra_compile_args)
-        + str(cffi_debug)
-        + sysconfig.get_config_var("CFLAGS")
-        + sysconfig.get_config_var("SOABI")
+        xstr(cffi_extra_compile_args) + xstr(cffi_debug)
+        # + sysconfig.get_config_var("CFLAGS")
```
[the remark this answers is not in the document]
jhale: Indeed, as mentioned the TODOs in the opening post - I have a list on win32 from @chrisrichardson and need to look through them.

minrk:
> is there a possibility option to use clang on Windows via CFFI?

clang is available on Windows, though I've never used it, so don't know what the implications would be. I believe if you use msvc, it would means folks need to install a compiler outside of conda. I _think_ clang should avoid this, but there may still be an SDK requirement, I'm not sure.

Question is: what packages would need to use clang:

- all fenics packages (basix, dolfinx)?
- none, just the runtime requirement?

It's going to take me a bit to figure out testing beyond basix on conda-forge, since the dependency means the unreleased builds need to be published somewhere that's not the conda-forge default channel, so downstream can depend on them.

jhale:
> > is there a possibility option to use clang on Windows via CFFI?
> 
> clang is available on Windows, though I've never used it, so don't know what the implications would be. I believe if you use msvc, it would means folks need to install a compiler outside of conda. I _think_ clang should avoid this, but there may still be an SDK requirement, I'm not sure.
> 
> Question is: what packages would need to use clang:
> 
> * all fenics packages (basix, dolfinx)?
> * none, just the runtime requirement?
> 
> It's going to take me a bit to figure out testing beyond basix on conda-forge, since the dependency means the unreleased builds need to be published somewhere that's not the conda-forge default channel, so downstream can depend on them.

I took a look at how FFCx is currently using `cffi.compile()` and on Windows MSVC is the only option without implementing our own JIT compiler wrappers (which I don't want to do, unless it is absolutely necessary) - so users will need to install MSVC. Most of our users will be covered by the Community License anyway.

This issue (how can I use a different compiler with cffi?) is discussed here:

https://foss.heptapod.net/pypy/cffi/-/issues/560

Our only runtime dependency on a compiler is when using CFFI via FFCx, but it is an important core feature and cannot be skipped.

https://github.com/conda-forge/cffi-feedstock/blob/main/recipe/meta.yaml

You can see here there is no runtime dependency on a C compiler, so it must be assumed provided by the system.

jhale:
On the last point, @chrisrichardson has a Windows machine setup for development, so perhaps we could work together to try a conda build locally?

minrk:
I [asked](https://conda.discourse.group/t/runtime-requirement-for-c-compiler-on-windows/614/3), and even with clang, users would need Windows SDKs to compile, so there would be little benefit to depending on clang instead of msvc.

jhale:
Thanks! I read the thread - we can revisit clang as an option at a later date. For now MSVC2022 using `cl.exe` will be the only officially supported options on Windows.

jhale:
Just to clarify a few points from https://conda.discourse.group/t/runtime-requirement-for-c-compiler-on-windows/614/3
- FFCx has a runtime dependency on a C (not C++) compiler only via CFFI.
- Our demo runner for testing has a dependency on a C compiler, but that is not necessary for a user install.
- We do not link with anything other than the standard links (stdc) etc. and libm (only on Linux platforms).

ffcx/codegeneration/jit.py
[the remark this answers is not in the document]
jhale: Options were pretty limited on Windows but we found `EXT_SUFFIX` with `.cp312-win_amd64.pyd`.

minrk:
Thanks! That clarifies things.

jhale:
Great. This patch is now ready for final review and preliminary testing in Conda.

minrk:
Haven't quite worked out all the kinks in getting basix from my channel into conda-forge CI, but so far looks like ffcx works fine on Windows in conda-forge with no modifications.

demo/test_demos.py
```diff
@@ -38,16 +42,33 @@ def test_demo(file, scalar_type):
...
+    else:
+        opts = f"--scalar_type {scalar_type}"
+        extra_flags = (
+            "-std=c17 -Wunused-variable -Werror -fPIC -Wno-error=implicit-function-declaration"
+        )
+        assert os.system(f"cd {demo_dir} && ffcx {opts} {file}.py") == 0
+        assert (
+            os.system(
+                f"cd {demo_dir} && "
+                f"gcc -I../ffcx/codegeneration "
```
garth-wells, line 69:
```suggestion
                f"cc -I../ffcx/codegeneration "
```

minrk: fwiw, I think `cc = os.environ.get("CC", sysconfig.get_config_var("CC"))` is more likely to get the right thing and accepts the ubiquitous standard env override.

minrk:
All tests pass on conda on Windows without any special handling: https://github.com/conda-forge/fenics-ffcx-feedstock/pull/7

Windows folks should be able to test installation with conda with:

```
conda install -c conda-forge -c minrk/label/fenics-windows fenics-ffcx=0.9.0.dev
```

jhale:
This is now ready to merge.

.github/workflows/pythonapp.yml
```diff
@@ -22,40 +22,75 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         python-version: ['3.9', '3.10', '3.11', '3.12']
+        include:
+          - os: windows-latest
+            python-version: '3.11'
```
garth-wells, line 27: Any reason to test with 3.11 rather than 3.12?

.github/workflows/pythonapp.yml
```diff
@@ -64,10 +99,15 @@ jobs:
         # Use always() to always run this step to publish test results
         # when there are test failures
         if: always()
+
+      - name: Setup cl.exe (Windows)
```
garth-wells, line 103: Is `cl.exe` clang?

demo/test_demos.py
```diff
@@ -17,6 +17,10 @@
 @pytest.mark.parametrize("scalar_type", ["float64", "float32", "complex128", "complex64"])
 def test_demo(file, scalar_type):
     """Test a demo."""
+    if sys.platform.startswith("win32") and "complex" in scalar_type:
+        # Skip complex demos on win32
+        pytest.skip(reason="_Complex not supported on Windows")
```
garth-wells, line 22: What a pain . . . .

garth-wells (commented):
Nice!

.github/workflows/pythonapp.yml
jhale: Not really, just providing some coverage because I will run against 3.12 on Windows elsewhere, e.g. DOLFINx.

.github/workflows/pythonapp.yml
jhale: No, it's Microsoft's Visual C Compiler (MSVC).

Note that Microsoft also offer `clang-cl.exe` which is a wrapper over `clang` to a provide drop-in replacement for `cl.exe`. This looked exciting in the first instance, because it supports `_Complex`. However, `clang-cl.exe` is not supported by Conda or CFFI, where MSVC is the only supported option today on Windows.

I did however add a test of `clang-cl.exe` to our demo pipeline, and it works fine.

demo/test_demos.py
jhale: Indeed, MSVC's lack of `_Complex` support has been the only major issue with doing a straightforward port.

Issue #699, opened 2024-05-27
LiteralFloat with value `Zero`  not supported in compilation of integral


jorgensd:
MWE:
```python
import basix.ufl
import ufl

c_el = basix.ufl.element("Lagrange", "interval", 1, shape=(1,))
mesh = ufl.Mesh(c_el)
C = ufl.as_tensor([[0, 0], [0, 1]])

integrand = ufl.transpose(C)[0, 0]

dx = ufl.Measure("dx", domain=mesh)
J = integrand * dx
forms = [J]
```
Traceback
```bash
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/root/shared/ffcx/ffcx/__main__.py", line 14, in <module>
    sys.exit(main())
  File "/root/shared/ffcx/ffcx/main.py", line 75, in main
    code_h, code_c = compiler.compile_ufl_objects(
  File "/root/shared/ffcx/ffcx/compiler.py", line 118, in compile_ufl_objects
    code = generate_code(ir, options)
  File "/root/shared/ffcx/ffcx/codegeneration/codegeneration.py", line 48, in generate_code
    code_integrals = [integral_generator(integral_ir, options) for integral_ir in ir.integrals]
  File "/root/shared/ffcx/ffcx/codegeneration/codegeneration.py", line 48, in <listcomp>
    code_integrals = [integral_generator(integral_ir, options) for integral_ir in ir.integrals]
  File "/root/shared/ffcx/ffcx/codegeneration/C/integrals.py", line 42, in generator
    parts = ig.generate()
  File "/root/shared/ffcx/ffcx/codegeneration/integral_generator.py", line 168, in generate
    all_quadparts += self.generate_quadrature_loop(rule)
  File "/root/shared/ffcx/ffcx/codegeneration/integral_generator.py", line 273, in generate_quadrature_loop
    tensor_comp, intermediates_fw = self.generate_dofblock_partition(quadrature_rule)
  File "/root/shared/ffcx/ffcx/codegeneration/integral_generator.py", line 387, in generate_dofblock_partition
    block_quadparts, intermediate = self.generate_block_parts(
  File "/root/shared/ffcx/ffcx/codegeneration/integral_generator.py", line 506, in generate_block_parts
[...]

PR #701, opened 2024-05-28
Remove zero Lexpr check

jorgensd:
Closing #699


michalhabera (approved):
Looks good, just fix the docstring to not mention zero simplification. Maybe add an issue for future work that there is this zero simplification opportunity.

PR #706, opened 2024-06-11
Make vertex quadrature test run without userwarning

jorgensd:
Combined with: https://github.com/FEniCS/ufl/pull/293 there are no warnings thrown when running the tests.


garth-wells (approved):

jorgensd:
We decided for an alternative approach in UFL, ref: https://github.com/FEniCS/ufl/issues/294

PR #714, opened 2024-09-13
Add CUDA wrapper capability.

bpachev:
This pull request adds the optional capability to generate CUDA-wrapped versions of the tabulate tensor functions. The wrappers are called at runtime to get the modified source code for each tabulate tensor, which can then be passed to the NVIDIA runtime compiler. This was originally developed by James Trotter (https://www.sciencedirect.com/science/article/pii/S0167819123000571),  
and then modified to work with the current version of ffcx.

This capability supports assembly on GPUs, implemented here as an add-on package (https://github.com/bpachev/cuda-dolfinx).

Would love any feedback or suggestions on how to do this in a better way. Thanks!

### DIFF
--- a/ffcx/codegeneration/C/integrals.py
+++ b/ffcx/codegeneration/C/integrals.py
@@ -69,6 +69,13 @@ def generator(ir: IntegralIR, options):
     else:
         code["tabulate_tensor_complex64"] = ".tabulate_tensor_complex64 = NULL,"
         code["tabulate_tensor_complex128"] = ".tabulate_tensor_complex128 = NULL,"
+    if options.get("cuda"):
+        code["tabulate_tensor_cuda"] = (
+            f".tabulate_tensor_cuda = tabulate_tensor_cuda_{factory_name}"
+        )
+    else:
+        code["tabulate_tensor_cuda"] = ""
+
     np_scalar_type = np.dtype(options["scalar_type"]).name
     code[f"tabulate_tensor_{np_scalar_type}"] = (
         f".tabulate_tensor_{np_scalar_type} = tabulate_tensor_{factory_name},"
@@ -76,7 +83,7 @@ def generator(ir: IntegralIR, options):
 
     element_hash = 0 if ir.coordinate_element_hash is None else ir.coordinate_element_hash
 
-    implementation = ufcx_integrals.factory.format(
+    implementation = ufcx_integrals.get_factory(options).format(
         factory_name=factory_name,
         enabled_coefficients=code["enabled_coefficients"],
         enabled_coefficients_init=code["enabled_coefficients_init"],
@@ -89,6 +96,7 @@ def generator(ir: IntegralIR, options):
         tabulate_tensor_float64=code["tabulate_tensor_float64"],
         tabulate_tensor_complex64=code["tabulate_tensor_complex64"],
         tabulate_tensor_complex128=code["tabulate_tensor_complex128"],
+        tabulate_tensor_cuda=code["tabulate_tensor_cuda"],
     )
 
     return declaration, implementation
--- a/ffcx/codegeneration/C/integrals_template.py
+++ b/ffcx/codegeneration/C/integrals_template.py
@@ -30,9 +30,55 @@ ufcx_integral {factory_name} =
   {tabulate_tensor_float64}
   {tabulate_tensor_complex64}
   {tabulate_tensor_complex128}
+  {tabulate_tensor_cuda}
   .needs_facet_permutations = {needs_facet_permutations},
   .coordinate_element_hash = {coordinate_element_hash},
 }};
 
 // End of code for integral {factory_name}
 """
+
+cuda_wrapper = """
+
+// Begin CUDA wrapper for integral {factory_name}
+void tabulate_tensor_cuda_{factory_name}(int* num_program_headers,
+                                         const char*** program_headers,
+                                         const char*** program_include_names,
+                                         const char** out_program_src,
+                                         const char** tabulate_tensor_function_name)
+{{
+  const char* program_src = ""
+    "#define alignas(x)\\n"
+    "#define restrict __restrict__\\n"
+    "\\n"
+    "typedef unsigned char uint8_t;\\n"
+    "typedef unsigned int uint32_t;\\n"
+    "typedef double ufc_scalar_t;\\n"
+    "\\n"
+    "extern \\"C\\" __global__\\n"
+    "void tabulate_tensor_{factory_name}({scalar_type}* restrict A,\\n"
+    "                                    const {scalar_type}* restrict w,\\n"
+    "                                    const {scalar_type}* restrict c,\\n"
+    "                                    const {geom_type}* restrict coordinate_dofs,\\n"
+    "                                    const int* restrict entity_local_index,\\n"
+    "                                    const uint8_t* restrict quadrature_permutation\\n"
+    "                                    )\\n"
+    "{{\\n"
+    "{tabulate_tensor_quoted}\\n"
+    "}}";
+  *num_program_headers = 0;
+  *program_headers = NULL;
+  *program_include_names = NULL;
+  *out_program_src = program_src;
+  *tabulate_tensor_function_name = "tabulate_tensor_{factory_name}";
+}}
+
+// End CUDA wrapper for integral {factory_name}
+
+"""
+
+def get_factory(options):
+    if options.get("cuda"):
+        return cuda_wrapper + factory
+    else:
+        return factory
--- a/ffcx/codegeneration/ufcx.h
+++ b/ffcx/codegeneration/ufcx.h
@@ -125,6 +125,27 @@ extern "C"
       const uint8_t* restrict quadrature_permutation);
 #endif // __STDC_NO_COMPLEX__
 
+  /// Return CUDA C++ source code for the ufc_tabulate_tensor kernel
+  ///
+  /// @param[out] num_program_headers
+  ///         The number of headers required by the program
+  /// @param[out] program_headers
+  ///         Entire contents of each header file
+  /// @param[out] program_include_names
+  ///         Names of each header file
+  /// @param[out] program_src
+  ///         CUDA C++ source code for the program containing the
+  ///         tabulate_tensor function.
+  /// @param[out] tabulate_tensor_function_name
+  ///         The name of the device-side function.
+  ///
+  typedef void(ufcx_tabulate_tensor_cuda)(
+      int* num_program_headers,
+      const char*** program_headers,
+      const char*** program_include_names,
+      const char** program_src,
+      const char** tabulate_tensor_function_name);
+
   typedef struct ufcx_integral
   {
     const bool* enabled_coefficients;
@@ -134,6 +155,7 @@ extern "C"
     ufcx_tabulate_tensor_complex64* tabulate_tensor_complex64;
     ufcx_tabulate_tensor_complex128* tabulate_tensor_complex128;
 #endif // __STDC_NO_COMPLEX__
+    ufcx_tabulate_tensor_cuda* tabulate_tensor_cuda;
     bool needs_facet_permutations;
 
     /// Get the hash of the coordinate element associated with the geometry of the mesh.
--- a/ffcx/options.py
+++ b/ffcx/options.py
@@ -20,6 +20,7 @@ import numpy.typing as npt
 logger = logging.getLogger("ffcx")
 
 FFCX_DEFAULT_OPTIONS = {
+    "cuda": (bool, False, "generate CUDA wrapped versions of tabulate tensor functions", None),
     "epsilon": (float, 1e-14, "machine precision, used for dropping zero terms in tables.", None),
     "scalar_type": (
         str,
--- a/ffcx/codegeneration/jit.py
+++ b/ffcx/codegeneration/jit.py
@@ -68,6 +68,9 @@ UFC_INTEGRAL_DECL += "\n".join(
 UFC_INTEGRAL_DECL += "\n".join(
     re.findall(r"typedef void ?\(ufcx_tabulate_tensor_complex128\).*?\);", ufcx_h, re.DOTALL)
 )
+UFC_INTEGRAL_DECL += "\n".join(
+    re.findall(r"typedef void ?\(ufcx_tabulate_tensor_cuda\).*?\);", ufcx_h, re.DOTALL)
+)
 
 UFC_INTEGRAL_DECL += "\n".join(
     re.findall("typedef struct ufcx_integral.*?ufcx_integral;", ufcx_h, re.DOTALL)


ffcx/codegeneration/ufcx.h
```diff
@@ -134,6 +155,7 @@ extern "C"
     ufcx_tabulate_tensor_complex64* tabulate_tensor_complex64;
     ufcx_tabulate_tensor_complex128* tabulate_tensor_complex128;
 #endif // __STDC_NO_COMPLEX__
+    ufcx_tabulate_tensor_cuda* tabulate_tensor_cuda;
```
jhale, line 158: Should this be included via a macro only if a cuda compiler is used?

ffcx/options.py
```diff
@@ -20,6 +20,7 @@
 logger = logging.getLogger("ffcx")
 
 FFCX_DEFAULT_OPTIONS = {
+    "cuda": (bool, False, "generate CUDA wrapped versions of tabulate tensor functions", None),
```
jhale, line 23: Better strategy might be to always generate it and let the C preprocessor include it if CUDA is available.

jhale:
We should look at what CI options we have to keep this working long term.

bpachev:
CUDA Toolkit can be installed without a GPU on the device, so we can test if the generated CUDA code compiles without needing a GPU in CI.

ffcx/codegeneration/ufcx.h
bpachev: Right now tabulate_tensor_cuda contains a string with the tabulate tensor source code, which can be called at runtime to generate source code for NVRTC, so a cuda compiler isn't currently required for the ffcx output. . . .

jhale: I'm not an expert on CUDA so tell me if I am wrong.

So these wrappers are for runtime compilation.

I still think it's best practice to guard code which shouldn't be seen by certain compilers using macros. Does NVRTC define anything? I found https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html for NVCC.

On that note, is anything additionally possible with NVCC?

ffcx/options.py
jhale: This should clarify that they are NVRTC wrappers, rather than CUDA wrappers (which I associate with `nvcc`).

ffcx/codegeneration/ufcx.h
```diff
@@ -125,6 +125,27 @@ extern "C"
       const uint8_t* restrict quadrature_permutation);
 #endif // __STDC_NO_COMPLEX__
 
+  /// Return CUDA C++ source code for the ufc_tabulate_tensor kernel
```
jhale, line 128: Mention NVRTC

ffcx/codegeneration/C/integrals_template.py
```diff
@@ -30,9 +30,55 @@
...
   .needs_facet_permutations = {needs_facet_permutations},
   .coordinate_element_hash = {coordinate_element_hash},
 }};
 
 // End of code for integral {factory_name}
 """
+
+cuda_wrapper = """
+
+// Begin CUDA wrapper for integral {factory_name}
```
jhale, line 43: Mention NVRTC

ffcx/codegeneration/C/integrals.py
```diff
@@ -69,14 +69,21 @@ def generator(ir: IntegralIR, options):
     else:
         code["tabulate_tensor_complex64"] = ".tabulate_tensor_complex64 = NULL,"
         code["tabulate_tensor_complex128"] = ".tabulate_tensor_complex128 = NULL,"
+    if options.get("cuda"):
+        code["tabulate_tensor_cuda"] = (
+            f".tabulate_tensor_cuda = tabulate_tensor_cuda_{factory_name}"
```
jhale, line 74: Propose `tabulate_tensor_cuda_nvrtc` to differentiate with `tabulate_tensor_cuda` or ``tabulate_tensor_cuda_nvcc` in the future?

jhale:
In terms of testing infrastructure, we have:

https://github.com/FEniCS/ffcx/blob/main/demo/test_demos.py

but this is aimed at ahead of time compilers. Can we do something similar with NVRTC?

ffcx/codegeneration/ufcx.h
jhale: I've had a think about this and I see your point - the standard compiler compiles a C++ code which then calls NVRTC at runtime? In this case I simply think that needs clarification that this is all related to NVRTC.

ffcx/codegeneration/C/integrals_template.py
```diff
@@ -30,9 +30,55 @@
...
+    "                                    const {scalar_type}* restrict c,\\n"
+    "                                    const {geom_type}* restrict coordinate_dofs,\\n"
+    "                                    const int* restrict entity_local_index,\\n"
+    "                                    const uint8_t* restrict quadrature_permutation\\n"
+    "                                    )\\n"
+    "{{\\n"
+    "{tabulate_tensor_quoted}\\n"
+    "}}";
+  *num_program_headers = 0;
+  *program_headers = NULL;
```
jhale, line 70: Just a note that I think we use `math.h` in our generated functions.

ffcx/codegeneration/C/integrals_template.py
```diff
@@ -30,9 +30,55 @@
...
+                                         const char*** program_headers,
+                                         const char*** program_include_names,
+                                         const char** out_program_src,
+                                         const char** tabulate_tensor_function_name)
+{{
+  const char* program_src = ""
+    "#define alignas(x)\\n"
+    "#define restrict __restrict__\\n"
+    "\\n"
+    "typedef unsigned char uint8_t;\\n"
```
jhale, line 54: I saw Stackoverflow thread on this https://stackoverflow.com/questions/4189243/whats-a-good-alternative-to-uint8-t-when-it-is-not-provided-by-the-compiler; is it not possible to use `#include <stdint.h>`? These typedefs are likely to cause issues so it would be good to avoid them.

If it's unavoidable then we need a comment here with an explanation.

jhale (changes requested):

ffcx/codegeneration/ufcx.h
bpachev: That's right, the wrapper is compiled as C++ code. Then user code invokes NVRTC at runtime which processes the string returned by the wrapper. I'll go ahead and clarify this.

bpachev:
We should be able to do something similar with NVRTC, we'll just need to have an extra step where we compile using the regular C++ compiler, and then call the generated wrapper functions to get the input for NVRTC.

ffcx/codegeneration/C/integrals_template.py
bpachev: I believe NVRTC must include math by default - just tested a kernel with math functions to be sure, verified there were math functions in the code passed to NVRTC and things still functioned properly.

ffcx/codegeneration/C/integrals_template.py
bpachev: Unfortunately it looks like this is unavoidable. I've added a comment explaining the use of typedefs. Another relevant thread is https://stackoverflow.com/questions/50565200/including-c-standard-headers-in-cuda-nvrtc-code.

bpachev:
I've updated the PR with the requested renaming and explanatory comments. How should I go about adding CI for this? A separate PR after this is merged, or as part of this one?

drew-parsons:
More general question here, perhaps to motivate another PR later: cuda tends to be nvidia only, and closed source, unless that's changed recently. How much work would it take to enable other GPUs with open source tools (sycl, ROCm) ?

bpachev:
On the ffcx side with the current wrapper approach, honestly not too much. Downstream in the GPU assembly routines more refactoring effort will be required. This is my next development priority along with support for multi-GPU assembly.

garth-wells:
I have reservations about this change. We know from our own work that the standard FFCx tabulate kernels are not performant on GPUs. I don't want a change that encourages slow/sub-optimal code.

We should looks at how FFCx can be made extensible to allow user-modified generated code.

bpachev:
Garth, thanks for weighing in on this. I would love to see FFCx have easier options for customization. I agree that assembly with high-order FFCx tabulate kernels is not efficient on GPUs. However, in my testing, it's quite performant for order-1 and even order-2 elements, which cover a large percentage of use cases. The rationale behind this change is to start getting people to use FEniCS with GPU acceleration, which will hopefully spur development and result in delivering a mature and optimized GPU capability. 

That said, in the short term, all I really need is a way to map the generated body of an FFCx tabulate tensor to the corresponding ufcx_form object in C++. Thoughts about how to best implement that?

bpachev:
To provide some specific examples of performance, using Lagrange elements for assembly of a Poisson mass matrix results in 70% memory throughput on an NVIDIA GH200 with a cubic tetrahedral mesh (6M elements). If the form has enough terms, it can also get a good compute throughput. This is the case for applying the Streamline Upwind Petrov-Galerkin (SUPG) method to the shallow water equations, which resulted in a 60% compute throughput for Jacobian matrix assembly on an NVIDIA A100 (also Lagrange, but triangular elements).

These represent pretty high levels of GPU resource utilization. The speedups relative to parallel CPU assembly are correspondingly high.

jhale:
I've taken another look at the PR and the comments from @garth-wells and @bpachev.

I think @garth-wells is right in that we should be careful to not encourage the early use of features that are potentially sub-optimal.

However this patch adds an optional low-level interface to a component of FEniCS (FFCx) that is essentially not user facing. Additionally, there is little scope to implement this differently. So, I see this as orthogonal to the path suggested by @garth-wells which is adding the capability to customise the generated code - after all, we would still want a well-defined interface in order to define how that customised code should be executed.

My remaining major with this PR is that we need a basic unit test to ensure that it continues working. We have this already for the standard C kernels, can this be extended/adapted?

https://github.com/FEniCS/ffcx/blob/main/demo/test_demos.py

@drew-parsons HIP has an equivalent RTC infrastructure:

https://rocm.docs.amd.com/projects/HIP/en/docs-6.0.0/user_guide/hip_rtc.html

bpachev:
@jhale Appreciate you taking another look at this. While I don't think we will be able to execute generated CUDA kernels without a GPU in the CI, we should certainly be able to test that the generated code compiles under NVRTC without errors. I'll update the PR with a test.

bpachev:
Ok. I've added a simple unit test to test_demo.py that tests compilation of a sample CUDA-wrapped tabulate tensor function with NVRTC.

ffcx/codegeneration/C/integrals.py
```diff
@@ -60,6 +60,7 @@ def generator(ir: IntegralIR, options):
         code["enabled_coefficients"] = "NULL"
 
     code["tabulate_tensor"] = body
+    code["tabulate_tensor_quoted"] = body.replace("\n", '\\n"\n    "')
```
jhale, line 63: Is this always necessary to output? Will increase the code size a lot for large kernels.

demo/test_demos.py
```diff
@@ -73,3 +73,43 @@ def test_demo(file, scalar_type):
...
+                f"{extra_flags} "
+                f"-c {file}.c"
+            )
+            == 0
+        )
+        cxx = os.environ.get("CXX", "c++")
+        assert (
+            os.system(
+                f"cd {demo_dir} && "
+                f"{cxx} -I../ffcx/codegeneration -I{nvrtc_dir}/include -L{nvrtc_dir}/lib"
```
jhale, line 107: Can we build with a higher C++ standard? e.g. 20.

And also -Werror

demo/test_demos.py
```diff
@@ -73,3 +73,43 @@ def test_demo(file, scalar_type):
...
+
+@pytest.mark.parametrize("scalar_type", ["float64", "float32"])
+def test_demo_nvrtc(scalar_type):
+    """Test generated CUDA code with NVRTC."""
+    file = "Components"
+    opts = f"--scalar_type {scalar_type} --cuda"
+    if not sys.platform.startswith("linux"):
+        pytest.skip(reason="NVRTC support only tested on Linux")
+    else:
+        from nvidia import cuda_nvrtc
```
jhale, line 86: We need an import skip on this test if cuda_nvrtc doesn't exist

Remove platform restrictions - I see no reason to artificially restrict this to Linux.

demo/nvrtc_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+          std::string(nvrtcGetErrorString(nvrtc_err))) + " "
+          "at " + std::string(__FILE__) + ":" + std::to_string(__LINE__);
+      }
+      if (log_size > 0)
+        program_log.resize(log_size-1);
+    }
+    nvrtcDestroyProgram(&program);
+
+    std::stringstream ss;
+    ss << "nvrtcCompileProgram() failed with "
```
jhale, line 75: You can use std::format if you can switch to C++20, removes the need for using C++ terrible string formatting.

demo/nvrtc_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+  // Compile the CUDA C++ program
+  nvrtcResult nvrtc_compile_err = nvrtcCompileProgram(
+    program, num_compile_options, compile_options);
+  if (nvrtc_compile_err != NVRTC_SUCCESS) {
+    // If the compiler failed, obtain the compiler log
+    std::string program_log;
+    size_t log_size;
+    nvrtc_err = nvrtcGetProgramLogSize(program, &log_size);
+    if (nvrtc_err != NVRTC_SUCCESS) {
+      program_log = std::string(
```
jhale, line 55: Std::format?

demo/nvrtc_test.cpp
```diff
@@ -0,0 +1,89 @@
...
+        "nvrtcGetProgramLogSize() failed with " +
+        std::string(nvrtcGetErrorString(nvrtc_err)) + " "
+        "at " + std::string(__FILE__) + ":" + std::to_string(__LINE__));
+    } else {
+      program_log.resize(log_size);
+      nvrtc_err = nvrtcGetProgramLog(
+        program, const_cast<char*>(program_log.c_str()));
+      if (nvrtc_err != NVRTC_SUCCESS) {
+        program_log = std::string(
```
jhale, line 64: Std::format?

ffcx/options.py
jhale, line 23: Again let's be more specific here and say nvrtc.

jhale (changes requested):

demo/test_demos.py
```diff
@@ -73,3 +73,43 @@ def test_demo(file, scalar_type):
             )
             == 0
         )
+
+
+@pytest.mark.parametrize("scalar_type", ["float64", "float32"])
+def test_demo_nvrtc(scalar_type):
+    """Test generated CUDA code with NVRTC."""
+    file = "Components"
```
jhale, line 81: What is this? Can we run across more kernels? Would be good to cover at least a few standard ones, poisson, elasticity etc.

demo/test_demos.py
bpachev: So currently the cuda_nvrtc package won't install on MacOS, which is the rationale for restriction to Linux. This is also why I stuck the package under optional dependencies, because those are only installed in the CI for Linux.

demo/nvrtc_test.cpp
bpachev: So I tried switching this code to use std::format, however the C++ compiler in Github's CI environment appears to not support c++20 (tests worked locally, but broke during CI on Github).

pyproject.toml
```diff
@@ -32,7 +32,7 @@ ffcx = "ffcx:__main__.main"
 [project.optional-dependencies]
 lint = ["ruff"]
 docs = ["sphinx", "sphinx_rtd_theme"]
-optional = ["numba", "pygraphviz==1.7"]
+optional = ["numba", "pygraphviz==1.7", "nvidia-cuda-nvrtc-cu12"]
 test = ["pytest >= 6.0", "sympy", "numba"]
```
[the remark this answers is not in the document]
bpachev: In the current CI, the optional dependencies are installed only on Linux, while the others are installed on all platforms. Unfortunately, sticking nvidia rtc in the test dependencies breaks the MacOS CI. Ideally we'd support all platforms, but I don't want to add CUDA Toolkit as a dependency and for the time being testing on Linux should suffice to ensure correctness of the generated CUDA wrapper code.

bpachev:
Ok, @jhale I've updated the PR to test with a wider range of forms, addressed the issues with tabulate_tensor_quoted, and renamed the --cuda option to --cuda_nvrtc. 

The nvrtc pip package does install on Windows (but not MacOS), although I haven't been able to confirm if it works the same (I don't have a Windows dev env, and have been testing via Github's CI). Futhermore, the C++ compiler on Windows has a pretty strict limit on the maximum character count for a string literal, which is a problem when wrapping the tensor for the elasticity demo, as it has a line over that limit.

The current PR only explicitly skips Windows in the test, but the import skip will effectively skip both MacOS and Windows. Thanks again for giving this PR a second chance and for your feedback!

demo/test_demos.py
bpachev: I've added a few more kernels.

ffcx/options.py
bpachev: Agreed

ffcx/codegeneration/C/integrals.py
bpachev: This only becomes part of the generated code if --cuda is set. Otherwise the format parameter is not used. However, I've updated the code to make the format parameter an empty string if --cuda is not set.

demo/test_demos.py
bpachev: Added -Werror but had issues in CI with C++20

jhale:
I'll get this finished and merged once the UFL typing stuff is fixed (not related to your PR) - I don't have time at the moment to do it myself.

PR #717, opened 2024-10-03
Remove quadrature permutation on element tables of codim-1

jorgensd:
The input already has a consistent orientation due to: https://github.com/FEniCS/dolfinx/pull/3209

### DIFF
--- a/ffcx/ir/elementtables.py
+++ b/ffcx/ir/elementtables.py
@@ -366,7 +366,7 @@ def build_optimized_tables(
         # differently to their global orientation
         if integral_type == "interior_facet" or (is_mixed_dim and codim == 0):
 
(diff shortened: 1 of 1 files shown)


ffcx/ir/elementtables.py
```diff
@@ -365,7 +365,10 @@ def build_optimized_tables(
         # needed because a cell may see its sub-entities as being oriented
         # differently to their global orientation
         if integral_type == "interior_facet" or (is_mixed_dim and codim == 0):
-            if tdim == 1:
+
+            if tdim == 1 or codim ==1:
```
jorgensd, line 369:
```suggestion
            if tdim == 1 or codim == 1:
```

PR #725, opened 2024-10-14
Add many geometrical quantities to ffcx expression generator.

jorgensd:
Fix various errors in the definition of geometry tables

### DIFF
--- a/ffcx/codegeneration/expression_generator.py
+++ b/ffcx/codegeneration/expression_generator.py
@@ -61,10 +61,16 @@ class ExpressionGenerator:
 
     def generate_geometry_tables(self):
         """Generate static tables of geometry data."""
-        # Currently we only support circumradius
         ufl_geometry = {
+            ufl.geometry.FacetEdgeVectors: "facet_edge_vertices",
+            ufl.geometry.CellFacetJacobian: "reference_facet_jacobian",
             ufl.geometry.ReferenceCellVolume: "reference_cell_volume",
+            ufl.geometry.ReferenceFacetVolume: "reference_facet_volume",
+            ufl.geometry.ReferenceCellEdgeVectors: "reference_edge_vectors",
+            ufl.geometry.ReferenceFacetEdgeVectors: "facet_reference_edge_vectors",
+            ufl.geometry.FacetJacobianDeterminant: "reference_facet_jacobian",
             ufl.geometry.ReferenceNormal: "reference_facet_normals",
+            ufl.geometry.FacetOrientation: "facet_orientation",
         }
 
         cells: dict[Any, set[Any]] = {t: set() for t in ufl_geometry.keys()}  # type: ignore
--- a/ffcx/codegeneration/expression_generator.py
+++ b/ffcx/codegeneration/expression_generator.py
@@ -68,9 +68,7 @@ class ExpressionGenerator:
             ufl.geometry.ReferenceFacetVolume: "reference_facet_volume",
             ufl.geometry.ReferenceCellEdgeVectors: "reference_edge_vectors",
             ufl.geometry.ReferenceFacetEdgeVectors: "facet_reference_edge_vectors",
-            ufl.geometry.FacetJacobianDeterminant: "reference_facet_jacobian",
             ufl.geometry.ReferenceNormal: "reference_facet_normals",
-            ufl.geometry.FacetOrientation: "facet_orientation",
         }
 
         cells: dict[Any, set[Any]] = {t: set() for t in ufl_geometry.keys()}  # type: ignore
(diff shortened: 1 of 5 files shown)


ffcx/codegeneration/expression_generator.py
```diff
@@ -61,10 +61,16 @@ def generate(self):
 
     def generate_geometry_tables(self):
         """Generate static tables of geometry data."""
-        # Currently we only support circumradius
         ufl_geometry = {
+            ufl.geometry.FacetEdgeVectors: "facet_edge_vertices",
```
michalhabera, line 65: Ugh, the naming (unrelated to the PR) is a bit inconsistent with UFL. Nvm, can you add a quick test so we're sure these are covered for the future?

jorgensd: Yep, I'll try to add a test sometime this week.

jorgensd: @michalhabera what is the best way to test these? It is a pain to derive all these expressions by hand.

As 99 % of these expressions rely on basix, alot of tests will just check that we call basix in a consistent fashion.
Is this sufficient?

jorgensd:
@michalhabera I've added tests that uses basix. Thus this checks that our expression generator handles the basix data correctly.

michalhabera:
> @michalhabera I've added tests that uses basix. Thus this checks that our expression generator handles the basix data correctly.

I think that your test logic is correct. We basix evaluates these we should not test against numerical values, but against what basix returns.

But can you unify naming there for translated names? E.g. `reference_facet_jacobian` -> `facet_jacobians`, so we translate UFL names to basix names. Currently there is additional naming scheme for which I do not see the point and system.

jorgensd:
> > @michalhabera I've added tests that uses basix. Thus this checks that our expression generator handles the basix data correctly.
> 
> I think that your test logic is correct. We basix evaluates these we should not test against numerical values, but against what basix returns.
> 
> But can you unify naming there for translated names? E.g. `reference_facet_jacobian` -> `facet_jacobians`, so we translate UFL names to basix names. Currently there is additional naming scheme for which I do not see the point and system.

Done

ffcx/codegeneration/geometry.py
```diff
@@ -121,7 +121,7 @@ def facet_reference_edge_vectors(tablename, cellname):
     return L.ArrayDecl(symbol, values=out, const=True)
 
 
-def reference_facet_normals(tablename, cellname):
+def reference_normals(tablename, cellname):
```
michalhabera, line 124: This would be `facet_outward_normals`.

test/test_jit_expression.py
```diff
@@ -323,3 +323,139 @@ def test_facet_expression(compile_args):
...
+    ffi_coords = ffi.cast(f"{c_xtype} *", coords.ctypes.data)
+    ffi_ei = ffi.cast("int *", entity_index.ctypes.data)
+    ffi_qp = ffi.cast("uint8_t *", quad_perm.ctypes.data)
+
+    # Check CellFacetJacobian
+    obj_fj = ffcx.codegeneration.jit.compile_expressions(
+        [(ufl.geometry.CellFacetJacobian(mesh), points)], cffi_extra_compile_args=compile_args
+    )[0][0]
+    output = np.zeros((2, 1), dtype=dtype)
+    obj_det = ffcx.codegeneration.jit.compile_expressions(
```
michalhabera, line 356: Why is this tested? FacetJacobianDeterminant is lowered already in the UFL

ffcx/codegeneration/geometry.py
jorgensd: Then we end up with a mixture of `basix` or `ufl`-style nomenclature.
Isn't it better to stick with UFL all the way through (as basix doesn't have concepts such as reference_facet_edge_vectors).

test/test_jit_expression.py
```diff
@@ -323,3 +323,139 @@ def test_facet_expression(compile_args):
...
+        [(ufl.geometry.FacetJacobianDeterminant(mesh), points)],
+        cffi_extra_compile_args=compile_args,
+    )[0][0]
+    output_det = np.zeros(1, dtype=dtype)
```
jorgensd, line 360:
```suggestion
```

test/test_jit_expression.py
```diff
@@ -323,3 +323,139 @@ def test_facet_expression(compile_args):
...
+        np.testing.assert_allclose(output, ref_fj)
+
+        obj_det.tabulate_tensor_float64(
+            ffi.cast(f"{c_type} *", output_det.ctypes.data),
+            ffi_coeff,
+            ffi_const,
+            ffi_coords,
+            ffi_ei,
+            ffi_qp,
+        )
```
jorgensd, line 382:
```suggestion
```

ffcx/codegeneration/geometry.py
michalhabera: Ah, I see, I wrongly thought you went for basix naming. All UFL-like works too 👍 .

michalhabera (approved):

PR #730, opened 2024-11-22
Update windows and mac CI and updates for ruff

jorgensd:


jhale:
DOLFINx integration test fails due to mixed element issue, otherwise should be OK.

PR #731, opened 2024-12-05
Codim 2 coupling

jorgensd:
This PR introduces code-generation for 3D-1D/2D-0D coupling over "ridges" of a cell.

It follows the same path as the mixed dimensional assembly for 3D-2D/2D-1D as made by @jpdean.

Currently, we only support "exterior" ridge integrals, meaning integrals with a one-sided (cell) view of the ridge.
This is due to the lack of arbitrary restriction operators within UFL.

The code does not touch any of the DOLFINx functionality/API, which will be in a separate PR.

The PR also adds type hints in some appropriate places.


test/test_jit_forms.py
```diff
@@ -1224,3 +1225,220 @@ def tabulate_tensor(ele_type, V_cell_type, W_cell_type, coeffs):
...
+            for flips in edge_flips[local_entity_index]:
+                A_ref[:, flips] = A_ref[:, flips[::-1]]
+        A_ref[:] = A_ref[::-1]
+    # # If the permutation is 1, this means the triangle sees its edge as being flipped
+    # # relative to the edge's global orientation. Thus the result is the same as swapping
+    # # cols 1 and 2 and cols 4 and 5 of the reference result.
+    # if permutation[0] == 1:
+    #     A_ref[:, [1, 2]] = A_ref[:, [2, 1]]
+    #     A_ref[:, [4, 5]] = A_ref[:, [5, 4]]
+    # print(A.T)
```
jorgensd, line 1341: @jpdean COuld you have a look at this logic?

ffcx/codegeneration/C/form.py
```diff
@@ -87,7 +87,7 @@ def generator(ir: FormIR, options):
     integral_ids = []
     integral_offsets = [0]
     # Note: the order of this list is defined by the enum ufcx_integral_type in ufcx.h
-    for itg_type in ("cell", "exterior_facet", "interior_facet"):
+    for itg_type in ("cell", "exterior_facet", "interior_facet", "ridge"):
```
mscroggs, line 90: Are there problems where you may want to only include ridges on the boundary? If so, we should include both `exterior_ridge` and `interior_ridge` here.

ffcx/codegeneration/access.py
```diff
@@ -376,10 +388,10 @@ def facet_edge_vectors(self, mt, tabledata, num_points):
 
         # Get edge vertices
         facet = self.symbols.entity("facet", mt.restriction)
-        facet_edge = mt.component[0]
+        edge_ridge = mt.component[0]
```
mscroggs, line 391: Should this variable have changed? Looks like it may be a Find & Replace gone rogue

ffcx/ir/elementtables.py
```diff
@@ -200,7 +199,7 @@ def generate_psi_table_name(
     if any(derivative_counts):
         name += "_D" + "".join(str(d) for d in derivative_counts)
     name += {None: "", "cell": "_AC", "facet": "_AF"}[averaged]
-    name += {"cell": "", "facet": "_F", "vertex": "_V"}[entity_type]
+    name += {"cell": "", "facet": "_F", "vertex": "_V", "ridge": "_E"}[entity_type]
```
mscroggs, line 202: Would `_R` be better for ridges? `_E` suggests edges

ffcx/codegeneration/C/form.py
jorgensd: I think this is a multi-step approach. lets start with exterior ridges. 
Interior ridges have the issue that the restriction in `ufl` is not really sufficient (an edge can be connected to `n` cells, thus +, - is not sufficient).

test/test_jit_forms.py
```diff
@@ -1224,3 +1225,220 @@ def tabulate_tensor(ele_type, V_cell_type, W_cell_type, coeffs):
...
+    print()
+    print(A_ref.T)
+    print(A.T)
```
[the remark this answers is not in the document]
jorgensd: We need some guidance form @jpdean  on the transformation logic in the tests.

ffcx/ir/elementtables.py
jorgensd, line 202:
```suggestion
    name += {"cell": "", "facet": "_F", "vertex": "_V", "ridge": "_R"}[entity_type]
```

ffcx/codegeneration/C/form.py
mscroggs: Ah yes, I was mixing up +/- and boundary/internal. Ignore my comment

HoseongJeong:
Thank you so much to Dokken and all the developers. If it's not too much trouble, would it be possible for you to provide just one usage example?

jorgensd:
When CI passes, I will merge this PR, as it has sufficient testing, adds new functionality and adds some type-hints along the way as well.

jorgensd:
> Thank you very much to Dokken and all the developers. If it's not too much trouble, could you kindly provide just one usage example? As for the embedded ridge, it’s not yet available for use, correct?

There are two tests that illustrate how this an be used within FFCx. 
To get this to work with DOLFINx, an additional PR is required. That will probably happen next week.

ffcx/codegeneration/geometry.py
```diff
@@ -64,6 +66,14 @@ def cell_facet_jacobian(tablename, cellname):
     return L.ArrayDecl(symbol, values=out, const=True)
 
 
+def cell_ridge_jacobian(tablename, cellname):
+    """Write a reference facet jacobian."""
```
jorgensd:
```suggestion
    """Write a reference ridge jacobian."""
```

ffcx/ir/elementtables.py
```diff
@@ -316,36 +315,34 @@ def permute_quadrature_quadrilateral(points, reflections=0, rotations=0):
 def build_optimized_tables(
     quadrature_rule: QuadratureRule,
     cell: ufl.Cell,
-    integral_type: str,
+    integral_type: typing.Literal["interior_facet", "exterior_facet", "ridge", "cell"],
```
jorgensd:
```suggestion
    integral_type: typing.Literal["interior_facet", "exterior_facet", "ridge", "cell", "vertex"],
```

ffcx/ir/representation.py
```diff
@@ -320,10 +302,15 @@ def _compute_integral_ir(
                 # prescribed in certain cases.
 
                 degree = md["quadrature_degree"]
-                if integral_type != "cell":
+                if "facet" in integral_type:
```
jorgensd, line 305: @schnellerhase Does this change any logic for your vertex integrals?

schnellerhase: The quadrature scheme 'vertex' (averaging over vertex values and dividing by cell volume) has nothing to do with the integral type 'vertex'. Got confused with that quite a bit myself. Therefore this should have no effect has it only changes the quadrature rules.

PR #736, opened 2024-12-18
get vertex quadrature from basix

mscroggs:


ffcx/ir/representation.py
```diff
@@ -276,55 +275,11 @@ def _compute_integral_ir(
...
-                                1.0 / 8.0,
-                                1.0 / 8.0,
-                            ]
-                        ),
-                    )
-                else:
-                    raise RuntimeError(f"Vertex scheme is not supported for cell: {cellname}")
+                points = basix.cell.geometry(getattr(basix.CellType, cellname))
+                weights = np.array(
+                    [1.0 / points.shape[0] / basix.cell.volume(getattr(basix.CellType, cellname))]
```
michalhabera, line 280: Should this be multiplied with the reference cell volume instead? Surely integrating a constant 1.0 should return the cell volume.

mscroggs, line 280:
```suggestion
                    [basix.cell.volume(getattr(basix.CellType, cellname)) / points.shape[0]]
```
Well spotted, I misread the comment above

michalhabera (approved):
Much better. Maybe make an issue to add test that would check this codepath?

ffcx/ir/representation.py
```diff
@@ -276,55 +275,11 @@ def _compute_integral_ir(
...
-                            ]
-                        ),
-                    )
-                else:
-                    raise RuntimeError(f"Vertex scheme is not supported for cell: {cellname}")
+                points = basix.cell.geometry(getattr(basix.CellType, cellname))
+                weights = np.array(
+                    [basix.cell.volume(getattr(basix.CellType, cellname)) / points.shape[0]]
+                    * points.shape[0]
+                )
```
jorgensd, line 282:
```suggestion
                cell_volume = basix.cell.volume(getattr(basix.CellType, cellname)
                weights = np.full(points.shape[0], cell_volume / points.shape[0]], dtype=points.dtype)
```

mscroggs:
> Much better. Maybe make an issue to add test that would check this codepath?

Good idea: #738

jorgensd (approved):

PR #737, opened 2024-12-18
remove ufcx_integral_type

mscroggs:
Looks like it isn't used anywhere


mscroggs:
nevermind, it's used in DOLFINx

PR #751, opened 2025-02-23
Start adding some typing and documentation

jorgensd:
Start using `typing.Literal` for string-types, as there is so much implicit logic on the expected type.

Also add some doc-strings when needed.

### DIFF
--- a/ffcx/ir/elementtables.py
+++ b/ffcx/ir/elementtables.py
@@ -13,8 +13,11 @@ import numpy as np
 import numpy.typing as npt
 import ufl
 
+from ffcx.definitions import entity_types
 from ffcx.element_interface import basix_index
+from ffcx.ir.analysis.modified_terminals import ModifiedTerminal
 from ffcx.ir.representationutils import (
+    QuadratureRule,
     create_quadrature_points_and_weights,
     integral_type_to_entity_dim,
     map_integral_points,
@@ -82,7 +85,7 @@ def get_ffcx_table_values(
     integral_type,
     element,
     avg,
-    entity_type,
+    entity_type: entity_types,
     derivative_counts,
     flat_component,
     codim,
@@ -175,7 +178,12 @@ def get_ffcx_table_values(
 
 
 def generate_psi_table_name(
-    quadrature_rule, element_counter, averaged: str, entity_type, derivative_counts, flat_component
+    quadrature_rule: QuadratureRule,
+    element_counter,
+    averaged: str,
+    entity_type: entity_types,
+    derivative_counts,
+    flat_component,
 ):
     """Generate a name for the psi table.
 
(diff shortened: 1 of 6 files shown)


ffcx/ir/elementtables.py
```diff
@@ -500,9 +520,11 @@ def build_optimized_tables(
                 d = local_derivatives[i]
                 sub_tbl = j.tabulate(d, pts)[d]
                 sub_tbl = sub_tbl.reshape(1, 1, sub_tbl.shape[0], sub_tbl.shape[1])
-                for i in all_tensor_factors:
-                    if i.values.shape == sub_tbl.shape and np.allclose(i.values, sub_tbl):
-                        tensor_factors.append(i)
+                for tensor_factor in all_tensor_factors:
```
jorgensd, line 523: @IgorBaratta Here `i` was redeclared (from line 5.18 enumerate), I assume that was a bug.
Do you agree?

IgorBaratta: Yes, thanks for the fix

ffcx/ir/elementtables.py
```diff
@@ -293,26 +301,38 @@ def permute_quadrature_quadrilateral(points, reflections=0, rotations=0):
...
-    Input:
-      entity_type - str
-      modified_terminals - ordered sequence of unique modified terminals
-      FIXME: Document
+    Args:
+        quadrature_rule: The quadrature rule relating to the tables.
+        cell: The cell type of the domain the tables will be used with.
+        entity_type: On what entity (vertex,edge,facet,cell) the tables are evaluated at.
+        integral_type: The type of integral the tables are used for.
+        modified_terminals: ordered sequence of unique modified terminals
```
mscroggs, line 322:
```suggestion
        modified_terminals: Ordered sequence of unique modified terminals
```

ffcx/ir/elementtables.py
```diff
@@ -293,26 +301,38 @@ def permute_quadrature_quadrilateral(points, reflections=0, rotations=0):
...
     """Build the element tables needed for a list of modified terminals.
 
-    Input:
-      entity_type - str
-      modified_terminals - ordered sequence of unique modified terminals
-      FIXME: Document
+    Args:
+        quadrature_rule: The quadrature rule relating to the tables.
+        cell: The cell type of the domain the tables will be used with.
+        entity_type: On what entity (vertex,edge,facet,cell) the tables are evaluated at.
```
mscroggs, line 320:
```suggestion
        entity_type: The entity type (vertex,edge,facet,cell) that the tables are evaluated for.
```

ffcx/analysis.py
```diff
@@ -246,7 +248,7 @@ def _analyze_form(
 
 
 def _has_custom_integrals(
-    o: ufl.integral.Integral | ufl.classes.Form | list | tuple,
+    o: typing.Union[ufl.integral.Integral, ufl.classes.Form, list, tuple],
```
mscroggs, line 251: How was this not causing an error previously? Or is 3.9 not running on CI so we just weren't seeing the error?

Is it worth considering bumping the minimun version to 3.10 six months early so we can use `|`?

mscroggs (approved):
A few minor comments. Otherwise looks good

ffcx/analysis.py
jorgensd: I think it depends on what functions are typed. Mypy ignores that are typed (if they are within an untyped function signature).

PR #749, opened 2025-02-19
sclaus/add void to kernels

sclaus2:
Add void* user_data to tabulate_tensor kernel to allow for easier extensions of non-standard integration kernels such as runtime kernels.

PR #753, opened 2025-03-07
 Add void* to tabulate_tensor kernel

sclaus2:
add void* to integration kernels to pass custom data to the integration kernel (related to https://github.com/FEniCS/ffcx/pull/749 ). 
This is a new pull request using branches in main FEniCS repo to allow for dolfinx tests to be run in github actions.

### DIFF
--- a/ffcx/codegeneration/utils.py
+++ b/ffcx/codegeneration/utils.py
@@ -96,6 +96,7 @@ if numba is not None:
     @numba.extending.intrinsic
     def empty_void_pointer(typingctx):
         """Custom intrinsic to return an empty void* pointer.
+
         This function creates a void pointer initialized to null (0).
         This is used to pass a nullptr to the UFCx tabulate_tensor interface.
 
@@ -104,7 +105,7 @@ if numba is not None:
 
         Returns:
             A Numba signature and a code generation function that returns a void pointer.
-        """  # noqa: D205
+        """
 
         def codegen(context, builder, signature, args):
             null_ptr = context.get_constant(numba.types.voidptr, 0)
--- a/ffcx/codegeneration/utils.py
+++ b/ffcx/codegeneration/utils.py
@@ -98,6 +98,7 @@ if numba is not None:
         """Custom intrinsic to return an empty void* pointer.
         This function creates a void pointer initialized to null (0).
         This is used to pass a nullptr to the UFCx tabulate_tensor interface.
+
         Args:
             typingctx: The typing context.
 
@@ -125,7 +126,8 @@ if numba is not None:
             arr: The NumPy array to get the void pointer from.
 
         Returns:
-            A Numba signature and a code generation function that returns a void pointer to the array's data.
+            A Numba signature and a code generation function that returns a void pointer
+            to the array's data.
         """
         if not isinstance(arr, numba.types.Array):
             raise TypeError("Expected a NumPy array")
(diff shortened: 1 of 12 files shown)


sclaus2:
Note that I had to change the branch for dolfinx_ref in dolfinx-tests.yml

ffcx/codegeneration/utils.py
```diff
@@ -80,6 +85,79 @@ def numba_ufcx_kernel_signature(dtype: npt.DTypeLike, xdtype: npt.DTypeLike):
...
+        """Custom intrinsic to get a void* pointer from a NumPy array.
+
+        This function takes a NumPy array and returns a void pointer to the array's data.
+        This is used to pass custom data organised in a NumPy array
+        to the UFCx tabulate_tensor interface.
+
+        Args:
+            typingctx: The typing context.
+            arr: The NumPy array to get the void pointer from.
+            In a multi-dimensional NumPy array, the memory is laid out in a contiguous
```
jhale, line 128: There is a bit more complexity possible in practice - instead, I would simply refer the reader to https://numpy.org/doc/stable/reference/arrays.ndarray.html#internal-memory-layout-of-an-ndarray

ffcx/codegeneration/utils.py
```diff
@@ -80,6 +85,79 @@ def numba_ufcx_kernel_signature(dtype: npt.DTypeLike, xdtype: npt.DTypeLike):
...
+            block of memory in either row-major (C-style) or
+            column-major (Fortran-style) order.
+            By default, NumPy uses row-major order.
+
+        Returns:
+            A Numba signature and a code generation function that returns a void pointer
+            to the first element of the contiguous block of memory that stores the array's
```
jhale, line 135: Remove the row-major comment as there is more complexity here than that.

test/test_custom_data.py
```diff
@@ -0,0 +1,118 @@
...
+    from cffi import FFI
+
+    # Define custom tabulate tensor function in C with a struct
+    # Step 1: Define the function in C and set up the CFFI builder
+    ffibuilder = FFI()
+    ffibuilder.set_source(
+        "_cffi_kernelA",
+        r"""
+        typedef struct {
+            uint8_t size;
```
jhale, line 23: Would be safer to use `size_t` for an example - `uint8_t` is very undersized.

test/test_custom_data.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    # Step 2: Compile the C code
+    ffibuilder.compile(verbose=True)
+
+    # Step 3: Import the compiled library
+    from _cffi_kernelA import ffi, lib
+
+    # Define cell data
+    size = 2
```
jhale, line 73: Pull size directly out from the ndarray?

ffcx/codegeneration/utils.py
```diff
@@ -80,6 +85,79 @@ def numba_ufcx_kernel_signature(dtype: npt.DTypeLike, xdtype: npt.DTypeLike):
...
+    def get_void_pointer(typingctx, arr):
+        """Custom intrinsic to get a void* pointer from a NumPy array.
+
+        This function takes a NumPy array and returns a void pointer to the array's data.
+        This is used to pass custom data organised in a NumPy array
+        to the UFCx tabulate_tensor interface.
+
+        Args:
+            typingctx: The typing context.
+            arr: The NumPy array to get the void pointer from.
```
jhale, line 127: The NumPy array to get the void pointer to the first element from.

jhale (changes requested):

jhale:
DOLFINx upstream refs need reverting to main once DOLFINx patch is merged.

PR #764, opened 2025-05-09
Vertex integrals

schnellerhase:
Introduces vertex integral support for FFCx. This corresponds to the `ufl.dP` measure. 

Accompanying PR in dolfinx: https://github.com/FEniCS/dolfinx/pull/3726 (not depending on it)

### DIFF
--- a/.github/workflows/dolfinx-tests.yml
+++ b/.github/workflows/dolfinx-tests.yml
@@ -81,7 +81,7 @@ jobs:
           ctest -V --output-on-failure -R unittests
 
       - name: Install Python demo/test dependencies
-        run: python3 -m pip install --break-system-packages matplotlib numba pyamg pytest pytest-xdist "scipy<=1.15"
+        run: python3 -m pip install --break-system-packages matplotlib numba pyamg pytest pytest-xdist "scipy<=1.15" pyvista
       - name: Run DOLFINx Python unit tests
         run: python3 -m pytest -n auto dolfinx/python/test/unit
       - name: Run DOLFINx Python demos
--- a/.github/workflows/dolfinx-tests.yml
+++ b/.github/workflows/dolfinx-tests.yml
@@ -81,7 +81,7 @@ jobs:
           ctest -V --output-on-failure -R unittests
 
       - name: Install Python demo/test dependencies
-        run: python3 -m pip install --break-system-packages matplotlib numba pyamg pytest pytest-xdist "scipy<=1.15" pyvista
+        run: python3 -m pip install --break-system-packages matplotlib numba pyamg pytest pytest-xdist "scipy<=1.15" pyvista networkx
       - name: Run DOLFINx Python unit tests
         run: python3 -m pytest -n auto dolfinx/python/test/unit
       - name: Run DOLFINx Python demos
--- a/ffcx/codegeneration/integral_generator.py
+++ b/ffcx/codegeneration/integral_generator.py
@@ -182,7 +182,8 @@ class IntegralGenerator:
 
         # No quadrature tables for custom (given argument) or point
         # (evaluation in single vertex)
-        skip = ufl.custom_integral_types + ufl.measure.point_integral_types
+        # skip = ufl.custom_integral_types + ufl.measure.point_integral_types
+        skip = ufl.custom_integral_types
         if self.ir.expression.integral_type in skip:
             return parts
 
--- a/ffcx/codegeneration/integral_generator.py
+++ b/ffcx/codegeneration/integral_generator.py
@@ -180,8 +180,7 @@ class IntegralGenerator:
         """Generate static tables of quadrature points and weights."""
         parts: list[L.LNode] = []
 
-        # No quadrature tables for custom (given argument) or point
-        # (evaluation in single vertex)
+        # No quadrature tables for custom (given argument)
         skip = ufl.custom_integral_types
         if self.ir.expression.integral_type in skip:
             return parts
--- a/ffcx/codegeneration/integral_generator.py
+++ b/ffcx/codegeneration/integral_generator.py
@@ -182,7 +182,6 @@ class IntegralGenerator:
 
         # No quadrature tables for custom (given argument) or point
         # (evaluation in single vertex)
-        # skip = ufl.custom_integral_types + ufl.measure.point_integral_types
         skip = ufl.custom_integral_types
         if self.ir.expression.integral_type in skip:
             return parts
--- a/test/test_jit_forms.py
+++ b/test/test_jit_forms.py
@@ -1404,3 +1404,36 @@ def test_ds_prism(compile_args, dtype):
             ]
         ),
     )
+
+
+@pytest.mark.parametrize("dtype", [np.float64])
+def test_point_measure(compile_args, dtype):
+    domain = ufl.Mesh(basix.ufl.element("Lagrange", "interval", 1, shape=(1,)))
+
+    dP = ufl.Measure("dP")
+    x = ufl.SpatialCoordinate(domain)
+    F = x[0] * dP
+    forms = [F]
+    compiled_forms, module, code = ffcx.codegeneration.jit.compile_forms(
+        forms, options={"scalar_type": dtype}, cffi_extra_compile_args=compile_args, cffi_debug=True
+    )
+
+    ffi = module.ffi
+    form0 = compiled_forms[0]
+    assert form0.rank == 0
+    default_integral = form0.form_integrals[0]
+    for a, b in [(0, 1), (1, 0), (2, 0), (5, -2)]:
+        J = np.zeros(1, dtype=np.float64)
+        coords = np.array([a, 0.0, 0.0, b, 0.0, 0.0], dtype=np.float64)
+        e = np.array([0], dtype=np.int32)
+        kernel = getattr(default_integral, "tabulate_tensor_float64")
+        kernel(
+            ffi.cast("double *", J.ctypes.data),
+            ffi.NULL,
+            ffi.NULL,
+            ffi.cast("double *", coords.ctypes.data),
+            ffi.cast("int *", e.ctypes.data),
+            ffi.NULL,
+            ffi.NULL,
+        )
+        assert np.isclose(J[0], a)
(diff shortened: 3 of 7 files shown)


jorgensd:
I think this is strongly related to: https://github.com/FEniCS/ffcx/pull/731 at least in 2D.

test/test_jit_forms.py
```diff
@@ -1404,3 +1404,83 @@ def test_ds_prism(compile_args, dtype):
             ]
         ),
     )
+
+
+@pytest.mark.parametrize("gdim", [1, 2, 3])
```
michalhabera, line 1409:
```suggestion
@pytest.mark.parametrize("geometry", [("interval", 1), ("triangle", 2), ("tetrahedron", 3)])
```

test/test_jit_forms.py
```diff
@@ -1404,3 +1404,83 @@ def test_ds_prism(compile_args, dtype):
...
+            kernel(
+                ffi.cast(f"{'float' if dtype == np.float32 else 'double'} *", J.ctypes.data),
+                ffi.NULL,
+                ffi.NULL,
+                ffi.cast(f"{'float' if dtype == np.float32 else 'double'} *", coords.ctypes.data),
+                ffi.cast("int *", e.ctypes.data),
+                ffi.NULL,
+                ffi.NULL,
+            )
+            assert np.isclose(J[i], coords[3 * i], atol=1e-6 if dtype == np.float32 else 1e-12)
```
jorgensd, line 1485: `1e2*np.finfo(dtype).eps`
as for instance done in: https://github.com/jorgensd/dolfinx-external-operator/blob/82a92a6099ddfde746cb73122d28d456816aea9c/doc/demo/demo_plasticity_von_mises_pure_ufl.py#L136

test/test_jit_forms.py
```diff
@@ -1404,3 +1404,83 @@ def test_ds_prism(compile_args, dtype):
...
+                ffi.cast("int *", e.ctypes.data),
+                ffi.NULL,
+                ffi.NULL,
+            )
+            assert np.isclose(J[0], coords[3 * i], atol=1e-6 if dtype == np.float32 else 1e-12)
+
+
+@pytest.mark.parametrize("gdim", [1, 2, 3])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_point_measure_rank_1(compile_args, gdim, dtype):
```
michalhabera, line 1449: Param over rank, add rank=2 case.

ffcx/codegeneration/integral_generator.py
```diff
@@ -182,7 +182,8 @@ def generate_quadrature_tables(self, domain: basix.CellType):
 
         # No quadrature tables for custom (given argument) or point
         # (evaluation in single vertex)
-        skip = ufl.custom_integral_types + ufl.measure.point_integral_types
+        # skip = ufl.custom_integral_types + ufl.measure.point_integral_types
```
michalhabera, line 185:
```suggestion
```

michalhabera (changes requested):
Only minor changes.

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,82 @@ def test_ds_prism(compile_args, dtype):
...
+        F = x[0] * ufl.conj(v) * dP
+    else:
+        F = x[0] * u * ufl.conj(v) * dP
+
+    (form0,), module, _ = ffcx.codegeneration.jit.compile_forms(
+        [F], options={"scalar_type": dtype}, cffi_extra_compile_args=compile_args
+    )
+    assert form0.rank == rank
+
+    def np_to_str(type):
```
jorgensd: Should this be moved somewhere more central in the tests, as I presume this could be used elsewhere?

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,82 @@ def test_ds_prism(compile_args, dtype):
...
+    ffi = module.ffi
+    kernel = getattr(
+        form0.form_integrals[0],
+        f"tabulate_tensor_{np_to_str(dtype)}",
+    )
+
+    for a, b in np.array([(0, 1), (1, 0), (2, 0), (5, -2)], dtype=dtype):
+        coords = np.array([a, 0.0, 0.0, b, 0.0, 0.0, 0, 0, 0, 0, 0, 1], dtype=rdtype)
+
+        for vertex in range(gdim + 1):
```
jorgensd: Should we rather use basix to get how many vertices there are in a cell, as then this test could be extended to quads and hexahedrons (in theory, with some rework of the coordinates).

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,82 @@ def test_ds_prism(compile_args, dtype):
...
+        assert False
+
+    ffi = module.ffi
+    kernel = getattr(
+        form0.form_integrals[0],
+        f"tabulate_tensor_{np_to_str(dtype)}",
+    )
+
+    for a, b in np.array([(0, 1), (1, 0), (2, 0), (5, -2)], dtype=dtype):
+        coords = np.array([a, 0.0, 0.0, b, 0.0, 0.0, 0, 0, 0, 0, 0, 1], dtype=rdtype)
```
jorgensd, line 1479:
```suggestion
        # General geometry for simplices of gdim 1,2 and 3.
        # gdim 1: creates the interval (a, b)
        # gdim 2: creates the triangle with vertices (a, 0), (b, 0), (0,0)
        # gdim 3: creates the tetrahedron with vertices (a, 0, 0), (b, 0, 0), (0, 0, 0), (0, 0, 1)
        coords = np.array([a, 0.0, 0.0, b, 0.0, 0.0, 0, 0, 0, 0, 0, 1], dtype=rdtype)
```

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,82 @@ def test_ds_prism(compile_args, dtype):
...
+    dP = ufl.Measure("dP")
+    x = ufl.SpatialCoordinate(domain)
+    V = ufl.FunctionSpace(domain, element)
+    u, v = ufl.TrialFunction(V), ufl.TestFunction(V)
+    if rank == 0:
+        F = x[0] * dP
+    elif rank == 1:
+        F = x[0] * ufl.conj(v) * dP
+    else:
+        F = x[0] * u * ufl.conj(v) * dP
```
jorgensd, line 1448:
```suggestion
    # With a Lagrange space for test and trial functions, all these forms should evaluate as the x-coordinate times an indicator for alignment between the argument(s) and the vertices .
    if rank == 0:
        F = x[0] * dP
    elif rank == 1:
        F = x[0] * ufl.conj(v) * dP
    else:
        F = x[0] * u * ufl.conj(v) * dP
```

test/test_jit_forms.py
schnellerhase: I found the other day `dtype().dtype.name` which does the same thing only more beautiful. Will remove.

.github/workflows/dolfinx-tests.yml
```diff
@@ -81,7 +81,7 @@ jobs:
           ctest -V --output-on-failure -R unittests
 
       - name: Install Python demo/test dependencies
-        run: python3 -m pip install --break-system-packages matplotlib numba pyamg pytest pytest-xdist "scipy<=1.15"
+        run: python3 -m pip install --break-system-packages matplotlib numba pyamg pytest pytest-xdist "scipy<=1.15" pyvista networkx
```
michalhabera, line 84: Are `pyvista` and `networkx` recently dependencies for dolfinx tests?

schnellerhase: Yes, pyvista not as a recent dependency, but recently changed to demo execution depending on it. networkx came with https://github.com/FEniCS/dolfinx/pull/3803

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,149 @@ def test_ds_prism(compile_args, dtype):
...
+                ffi.cast(f"{dtype_to_c_type(dtype)} *", J.ctypes.data),
+                ffi.NULL,
+                ffi.NULL,
+                ffi.cast(f"{dtype_to_c_type(rdtype)} *", coords.ctypes.data),
+                ffi.cast("int *", e.ctypes.data),
+                ffi.NULL,
+                ffi.NULL,
+            )
+            idx = (rank > 0) * vertex + (rank > 1) * vertex * vertex_count
+            assert np.isclose(J[idx], coords[vertex * 3], atol=1e2 * np.finfo(dtype).eps)
```
michalhabera:
```suggestion
            assert np.isclose(J[idx], coords[vertex * 3], rtol=1e2 * np.finfo(dtype).eps)
```
machine epsilon is the relative precision.

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,149 @@ def test_ds_prism(compile_args, dtype):
...
+            assert np.allclose(np.delete(J, [idx]), 0)
```
michalhabera, line 1496:
```suggestion
            # Check all other entries are not touched
            assert np.allclose(np.delete(J, [idx]), 0)
```
I think I'd add such comment for clarity.

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,149 @@ def test_ds_prism(compile_args, dtype):
...
+
+
+@pytest.mark.parametrize("geometry", [("interval", 1), ("triangle", 2), ("tetrahedron", 3)])
```
michalhabera: This is too much code and maintenance burden for checking that we raise. Either make it part of the first test or remove the test.

test/test_jit_forms.py
michalhabera: I see, fails when the true result is 0.0. Ok, I guess if this is fixed input of order 1 we can go back to atol, but proper general check should be more robust.

test/test_jit_forms.py
```diff
@@ -1404,3 +1405,92 @@ def test_ds_prism(compile_args, dtype):
...
+                ffi.cast(f"{dtype_to_c_type(dtype)} *", J.ctypes.data),
+                ffi.NULL,
+                ffi.NULL,
+                ffi.cast(f"{dtype_to_c_type(rdtype)} *", coords.ctypes.data),
+                ffi.cast("int *", e.ctypes.data),
+                ffi.NULL,
+                ffi.NULL,
+            )
+            idx = (rank > 0) * vertex + (rank > 1) * vertex * vertex_count
+            assert np.isclose(coords[vertex * 3], J[idx], atol=1e-6)
```
michalhabera: I mean the `atol` should still be 2 orders above machine eps, otherwise this is too loose for float64, just revert to what you had originally.

PR #768, opened 2025-06-26
Fix mypy for newly typed ufl

schnellerhase:
Unfortunately this means we need to ignore `attr-defined` errors now, since they get propagated up from ufl. Can be reverted with fixing of https://github.com/FEniCS/ufl/issues/386.

### DIFF
--- a/ffcx/analysis.py
+++ b/ffcx/analysis.py
@@ -34,7 +34,7 @@ class UFLData(typing.NamedTuple):
     # Lookup table from each unique element to its index in `unique_elements`
     element_numbers: dict[basix.ufl._ElementBase, int]
     # List of unique coordinate elements
-    unique_coordinate_elements: list[ufl.AbstractFiniteElement]
+    unique_coordinate_elements: list[basix.ufl._ElementBase]
     # List of ufl Expressions as tuples (expression, points, original_expression)
     expressions: list[tuple[ufl.core.expr.Expr, npt.NDArray[np.floating], ufl.core.expr.Expr]]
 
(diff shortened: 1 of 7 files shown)


ffcx/analysis.py
```diff
@@ -34,7 +34,7 @@ class UFLData(typing.NamedTuple):
     # Lookup table from each unique element to its index in `unique_elements`
     element_numbers: dict[basix.ufl._ElementBase, int]
     # List of unique coordinate elements
-    unique_coordinate_elements: list[basix.ufl._ElementBase]
+    unique_coordinate_elements: list[ufl.AbstractFiniteElement]
```
mscroggs: Is this change necessary? `basix.ufl._ElementBase` is a subclass of `ufl.AbstractFiniteElement` and FFCx probably doesn't work with non-Basix elements here

schnellerhase: Thanks, required some more consistency fixes, but should be done now.

ffcx/naming.py
```diff
@@ -18,7 +18,7 @@
 
 
 def compute_signature(
-    ufl_objects: list[ufl.Form | tuple[ufl.core.expr.Expr, npt.NDArray[np.float64]]],
+    ufl_objects: list[ufl.Form] | list[tuple[ufl.core.expr.Expr, npt.NDArray[np.floating]]],
```
schnellerhase, line 21:
```suggestion
    ufl_objects: list[ufl.Form | tuple[ufl.core.expr.Expr, npt.NDArray[np.floating]]],
```

PR #772, opened 2025-07-02
Fix 0D integrals.

jorgensd:
Reported at:
https://fenicsproject.discourse.group/t/block-form-fails-with-1d-mesh-and-hdg-scheme/17866

### DIFF
--- a/ffcx/ir/elementtables.py
+++ b/ffcx/ir/elementtables.py
@@ -255,7 +255,8 @@ def get_modified_terminal_element(mt) -> typing.Optional[ModifiedTerminalElement
     assert (mt.averaged is None) or not (ld or gd)
     # Change derivatives format for table lookup
     tdim = domain.topological_dimension()
-    local_derivatives: tuple[int, ...] = tuple(ld.count(i) for i in range(tdim))
+    num_derivatives = max(1, tdim)  # Handle 0-D domains, which for instance happens with sub-meshes
+    local_derivatives: tuple[int, ...] = tuple(ld.count(i) for i in range(num_derivatives))
 
     return ModifiedTerminalElement(element, mt.averaged, local_derivatives, fc)
 
@@ -461,6 +462,7 @@ def build_optimized_tables(
                     t = new_table[0]
                     t["array"] = np.vstack([td["array"] for td in new_table])
         else:
+            breakpoint()
             t = get_ffcx_table_values(
                 quadrature_rule.points,
                 cell,


ffcx/ir/elementtables.py
```diff
@@ -255,7 +255,8 @@ def get_modified_terminal_element(mt) -> typing.Optional[ModifiedTerminalElement
     assert (mt.averaged is None) or not (ld or gd)
     # Change derivatives format for table lookup
     tdim = domain.topological_dimension()
-    local_derivatives: tuple[int, ...] = tuple(ld.count(i) for i in range(tdim))
+    num_derivatives = max(1, tdim)  # Handle 0-D domains, which for instance happens with sub-meshes
```
garth-wells: Isn't the number of derivatives 0 for 0-D domains?

jorgensd: We are using `range(num_derivatives)` in the next line, meaning that we should include a `0`.
We can rename the variable though.

schnellerhase: That implies, that `local_derivatives` for both a 0D and 1D domain should have the same size of 1? Why is the (former) output of `local_derivatives` of `()` not correct here?

jorgensd: Because `ffcx` actually doesn't want the number of derivatives here.
Consider the MWE:
```python
import basix.ufl
import ufl

cell = basix.CellType.interval
c_el = basix.ufl.element("Lagrange", cell, 1, shape=(1,))
msh = ufl.Mesh(c_el)

el = basix.ufl.element("Lagrange", cell, 0, discontinuous=True)
V = ufl.FunctionSpace(msh, el)
u = ufl.Coefficient(V)
dx = ufl.Measure("dx", domain=msh)
J = u * dx
```
For this case the local deriviatives of the modified terminal is `()`, i.e. no derivatives.
Then the lines:
```python
    local_derivatives: tuple[int, ...] = tuple(ld.count(i) for i in range(domain.topological_dimension()))
```
will on `main` return `local_derivatives = (0, )`, as there are no derivatives in the list.

If we consider the same problem, but on a `point`/`vertex` mesh, i.e.
```python
import basix.ufl
import ufl

cell = basix.CellType.point
c_el = basix.ufl.element("Lagrange", cell, 0, shape=(1,), discontinuous=True)
msh = ufl.Mesh(c_el)

el = basix.ufl.element("Lagrange", cell, 0, discontinuous=True)
V = ufl.FunctionSpace(msh, el)
u = ufl.Coefficient(V)
dx = ufl.Measure("dx", domain=msh)
J = u * dx
```

we get `local_derivatives=()`, as we never enter the for-loop.

schnellerhase: I see , `local_derivates` originates from the argument of the terminal, not the coordinate element. And a 0D terminal is a constant so not what we want on the 0D domain.

schnellerhase (approved):

ffcx/ir/elementtables.py
garth-wells: > We are using `range(num_derivatives)` in the next line, meaning that we should include a `0`. We can rename the variable though.

Can you add some explanation to the code, and rename variables if appropriate? The code is confusing at the moment.

jorgensd: Explanation added. 
The summary is that `basix.index` does not support `0D` domains (whose index should always be `0`).
Maybe @mscroggs could comment further?

test/test_jit_forms.py
```diff
@@ -1407,6 +1407,7 @@ def test_ds_prism(compile_args, dtype):
     )
 
 
+
```
jorgensd:
```suggestion
```

PR #783, opened 2025-09-05
Diagonal assembly of matrices

jorgensd:
Crucial for matrix-free approaches with Jacobi smoothing.

The code let's the user pass in an extra options flag (`part`), which assemble into a rank-1 tensor (diagonal) instead of a rank-2 matrix (full).

The main alterations happens when:
- Gathering UFL forms, extracting the diagonal of mixed element problem
- the code generation, where diagonal matrices gets a loop removal.

Requires: https://github.com/FEniCS/ufl/pull/416

### DIFF
--- a/.github/workflows/dolfinx-tests.yml
+++ b/.github/workflows/dolfinx-tests.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Install UFL and Basix (default branches/tags)
         if: github.event_name != 'workflow_dispatch'
         run: |
-          python3 -m pip install --break-system-packages git+https://github.com/FEniCS/ufl.git
+          python3 -m pip install --break-system-packages git+https://github.com/FEniCS/ufl.git@dokken/diagonalise
           python3 -m pip install --break-system-packages git+https://github.com/FEniCS/basix.git
       - name: Install UFL and Basix (specified branches/tags)
         if: github.event_name == 'workflow_dispatch'
--- a/.github/workflows/dolfinx-tests.yml
+++ b/.github/workflows/dolfinx-tests.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Install UFL and Basix (default branches/tags)
         if: github.event_name != 'workflow_dispatch'
         run: |
-          python3 -m pip install --break-system-packages git+https://github.com/FEniCS/ufl.git@dokken/diagonalise
+          python3 -m pip install --break-system-packages git+https://github.com/FEniCS/ufl.git
           python3 -m pip install --break-system-packages git+https://github.com/FEniCS/basix.git
       - name: Install UFL and Basix (specified branches/tags)
         if: github.event_name == 'workflow_dispatch'
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -75,7 +75,7 @@ jobs:
         env:
           VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
         run: |
-          pip install git+https://github.com/FEniCS/ufl.git
+          pip install git+https://github.com/FEniCS/ufl.git@dokken/diagonalise
           pip install -v git+https://github.com/FEniCS/basix.git --config-settings=cmake.args=-DINSTALL_RUNTIME_DEPENDENCIES=ON --config-settings=cmake.args=-DCMAKE_TOOLCHAIN_FILE=C:/vcpkg/scripts/buildsystems/vcpkg.cmake
 
       - name: Install FFCx (Linux, with optional dependencies)
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -38,13 +38,13 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
 
       - name: Export GitHub Actions cache environment variables (Windows)
         if: runner.os == 'Windows'
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           script: |
             core.exportVariable('ACTIONS_CACHE_URL', process.env.ACTIONS_CACHE_URL || '');
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -67,7 +67,7 @@ jobs:
       - name: Install FEniCS dependencies (Python, Unix)
         if: runner.os == 'Linux' || runner.os == 'macOS'
         run: |
-          pip install git+https://github.com/FEniCS/ufl.git@dokken/diagonalise
+          pip install git+https://github.com/FEniCS/ufl.git
           pip install git+https://github.com/FEniCS/basix.git
 
       - name: Install FEniCS dependencies (Python, Windows)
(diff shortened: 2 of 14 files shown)


test/test_jit_forms.py
```diff
@@ -1848,3 +1848,79 @@ def tabulate_tensor(ele_type, V_cell_type, l_i):
...
+        ffi.cast(f"{c_type} *", c.ctypes.data),
+        ffi.cast(f"{c_xtype} *", coords.ctypes.data),
+        ffi.NULL,
+        ffi.NULL,
+        ffi.NULL,
+    )
+
+    compiled_forms, module, code = ffcx.codegeneration.jit.compile_forms(
+        forms, options={"scalar_type": dtype}, visualise=True, cffi_extra_compile_args=compile_args
+    )
```
jorgensd:
```suggestion
    compiled_forms, module, code = ffcx.codegeneration.jit.compile_forms(
        forms, options={"scalar_type": dtype}, cffi_extra_compile_args=compile_args
    )
```

test/test_jit_forms.py
```diff
@@ -1848,3 +1848,79 @@ def tabulate_tensor(ele_type, V_cell_type, l_i):
...
+    a = (
+        ufl.inner(u.dx(0), v.dx(2)) * ufl.dx
+        - ufl.inner(u, v) * ufl.dx
+        + ufl.inner(u.dx(0), v) * ufl.dx
+    )
+    forms = [a]
+    compiled_diag_forms, diag_module, _ = ffcx.codegeneration.jit.compile_forms(
+        forms,
+        options={"scalar_type": dtype, "diagonalise": True},
+        visualise=True,
```
jorgensd:
```suggestion
```

jhale:
Implementation is fine I think, just a comment on the interface.

Given lower and upper triangular are already being discussed, an enum or string argument called part with (default) full, diagonal, and eventually lower_triangular and upper_triangular might be better.

jhale:
Would have to be a string looking at ffcx options.

jorgensd:
> Implementation is fine I think, just a comment on the interface.
> 
> Given lower and upper triangular are already being discussed, an enum or string argument called part with (default) full, diagonal, and eventually lower_triangular and upper_triangular might be better.

I can implement it as an argument `part` which takes those strings, but will not implement the actual insertion of upper or lower diagonal, as it requires some careful consideration.

IgorBaratta:
LGTM! 
And I think we should handle upper and lower diagonal matrices in a separate PR.

Currently, FFCx doesn't have info about the local-to-global mapping. This means that even if a local matrix is an upper diagonal, it won't necessarily result in a global upper diagonal matrix.

IgorBaratta (approved):

jorgensd:
> Implementation is fine I think, just a comment on the interface.
> 
> Given lower and upper triangular are already being discussed, an enum or string argument called part with (default) full, diagonal, and eventually lower_triangular and upper_triangular might be better.

Rewritten the interface to use an Enum (StrEnum in Python 3.11) for the "part" of the matrix to assemble. Then we can handle the extension logic later.

jhale:
Right, no need to implement the others yet, just be good to have space to do so. Looks good.

ffcx/codegeneration/integral_generator.py
```diff
@@ -467,6 +476,18 @@ def generate_block_parts(
                 index = create_dof_index(table_ref, symbol)
                 B_indices.append(index)
 
+            if self.ir.part == TensorPart.diagonal and block_rank == 2:
+                assert len(A_shape) == 1
+                # If off-diagonal in mixed element, skip contribution
+                if B_indices[0].size() != B_indices[1].size():
```
michalhabera: I don't think this is the right way to check for off-diagonal mixed element contributions. You can have a mixed element with same number of dofs for each subelement, just consider MixedElement([P1, P1]), which will then incorrectly add contributions from mixed terms.

I've been looking at the code for some time now, but I cannot see a solution for this yet. One has the tabledata table name, but it is unfortunately not the right thing to compare, since e.g. two names `('FE3_C0_D100_Q929', 'FE3_C0_D001_Q929')` mean access to different tables, but might just come from the same diagonal argument product. In other words, we can now compare Modified Arguments, but more work is required to store the original Argument and compare blockdata based on that.

michalhabera (changes requested):

ffcx/codegeneration/integral_generator.py
jorgensd: The best thing would be to do it at the ufl level. I started an alternative version where I tried to use ufl.extract_blocks.

however, as the argument is replaced when using it, it is also not correct. 
one could make a copy of extract_blocks that doesn’t modify the arguments, and use it prior to analysis.

jorgensd: See: https://github.com/FEniCS/ufl/pull/416 for resolution.

test/test_jit_forms.py
```diff
@@ -1848,3 +1848,235 @@ def tabulate_tensor(ele_type, V_cell_type, l_i):
...
+        ffi.cast(f"{c_type} *", A_diag.ctypes.data),
+        ffi.cast(f"{c_type} *", w.ctypes.data),
+        ffi.cast(f"{c_type} *", c.ctypes.data),
+        ffi.cast(f"{c_xtype} *", coords.ctypes.data),
+        ffi.NULL,
+        ffi.NULL,
+        ffi.NULL,
+    )
+
+    compiled_forms, module, code = ffcx.codegeneration.jit.compile_forms(
```
jorgensd:
```suggestion
    compiled_forms, module, _code = ffcx.codegeneration.jit.compile_forms(
```

test/test_jit_forms.py
```diff
@@ -1848,3 +1848,235 @@ def tabulate_tensor(ele_type, V_cell_type, l_i):
...
```
jorgensd:
```suggestion
    compiled_forms, module, _code = ffcx.codegeneration.jit.compile_forms(
```

test/test_jit_forms.py
```diff
@@ -1848,3 +1848,235 @@ def tabulate_tensor(ele_type, V_cell_type, l_i):
...
```
jorgensd:
```suggestion
    compiled_forms, module, _code = ffcx.codegeneration.jit.compile_forms(
```
Last ruff compliance.

.github/workflows/dolfinx-tests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - name: Install UFL and Basix (default branches/tags)
         if: github.event_name != 'workflow_dispatch'
         run: |
-          python3 -m pip install --break-system-packages git+https://github.com/FEniCS/ufl.git
+          python3 -m pip install --break-system-packages git+https://github.com/FEniCS/ufl.git@dokken/diagonalise
```
jorgensd:
```suggestion
          python3 -m pip install --break-system-packages git+https://github.com/FEniCS/ufl.git
```

.github/workflows/pythonapp.yml
```diff
@@ -67,15 +67,15 @@ jobs:
       - name: Install FEniCS dependencies (Python, Unix)
         if: runner.os == 'Linux' || runner.os == 'macOS'
         run: |
-          pip install git+https://github.com/FEniCS/ufl.git
+          pip install git+https://github.com/FEniCS/ufl.git@dokken/diagonalise
```
jorgensd:
```suggestion
          pip install git+https://github.com/FEniCS/ufl.git
```

.github/workflows/pythonapp.yml
```diff
@@ -67,15 +67,15 @@ jobs:
...
           pip install git+https://github.com/FEniCS/basix.git
 
       - name: Install FEniCS dependencies (Python, Windows)
         if: runner.os == 'Windows'
         env:
           VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
         run: |
-          pip install git+https://github.com/FEniCS/ufl.git
+          pip install git+https://github.com/FEniCS/ufl.git@dokken/diagonalise
```
jorgensd:
```suggestion
          pip install git+https://github.com/FEniCS/ufl.git
```

PR #787, opened 2025-09-10
Introduce integral type and try to separate ufl concepts from FFCx concepts

jorgensd:
Following the discussion in: https://github.com/FEniCS/dolfinx/pull/3900
here follows an attempt of separate `UFL` concepts from `FFCx`/`DOLFINx` concepts.

This starts with replacing the `integral_type`, which we are just taking directly from ufl as a string, with a `NamedTuple` that contains:
- `codim`: An integer that is either 0, 1, ...., tdim to represent cells, facets, ridges and peaks, or -1 to represent vertices (for now as they are introduced in FFCx. I believe they should be replaced with peak integrals).
- `num_neighbours`: Number of neighbouring cells. This indicates if we have an "interior" or "exterior" quantity. Currently this is not used to do anything in UFCx.h, but i believe this is something we should store there, rather than the enum we are currently using (`facet`, `exterior_facet`, `cell`, `ridge`, `vertex`). Default value for any IntegralType is 1 neighbour
- `is_expression`: As the integral-ir of Expression and Integrands are the same, we store this information as an optional bool.
- `is_custom`: If anyone wants to implement a custom element, it should go through the custom measure (none of the other custom types) for now.

### DIFF
--- a/ffcx/ir/elementtables.py
+++ b/ffcx/ir/elementtables.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2013-2017 Martin Sandve Alnæs
+# Copyright (C) 2013-2025 Martin Sandve Alnæs and Jørgen S. Dokken
 #
 # This file is part of FFCx. (https://www.fenicsproject.org)
 #
@@ -13,7 +13,7 @@ import numpy as np
 import numpy.typing as npt
 import ufl
 
-from ffcx.definitions import entity_types
+from ffcx.definitions import IntegralType
 from ffcx.element_interface import basix_index
 from ffcx.ir.analysis.modified_terminals import ModifiedTerminal
 from ffcx.ir.representationutils import (
@@ -80,12 +80,11 @@ def clamp_table_small_numbers(
 
 
 def get_ffcx_table_values(
-    points,
-    cell,
-    integral_type,
-    element,
-    avg,
-    entity_type: entity_types,
+    points: npt.NDArray[np.floating],
+    cell: ufl.Cell,
+    integral_type: IntegralType,
+    element: basix.ufl._ElementBase,
+    avg: typing.Literal["cell", "facet", None],
     derivative_counts,
     flat_component,
     codim,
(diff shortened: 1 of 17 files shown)


ffcx/ir/representation.py
```diff
@@ -454,14 +450,20 @@ def _compute_form_ir(
     # Store names of integrals and subdomain_ids for this form, grouped
     # by integral types since form points to all integrals it contains,
     # it has to know their names for codegen phase
-    ufcx_integral_types = ("cell", "exterior_facet", "interior_facet", "vertex", "ridge")
+    ufcx_integral_types = (
+        IntegralType(codim=0),
+        IntegralType(codim=1, num_neighbours=1),
+        IntegralType(codim=1, num_neighbours=2),
+        IntegralType(codim=2),
+        IntegralType(codim=-1),
```
jorgensd:
```suggestion
        IntegralType(codim=-1),
        IntegralType(codim=2),
```

ffcx/ir/elementtables.py
```diff
@@ -97,18 +96,14 @@ def get_ffcx_table_values(
...
-        # or points on a facet
-        if entity_type == "cell":
-            integral_type = "cell"
-        else:
-            integral_type = "exterior_facet"
+    match integral_type:
+        case IntegralType(is_custom=True):
+            assert integral_type.codim == 0
+            integral_type = IntegralType(codim=0)
+            assert not avg
```
jorgensd:
```suggestion
    match integral_type:
        case IntegralType(is_custom=True):
            # Use quadrature points on cell for analysis in custom integral types
            assert integral_type.codim == 0
            integral_type = IntegralType(codim=0)
            assert not avg
```

ffcx/ir/elementtables.py
```diff
@@ -97,18 +96,14 @@ def get_ffcx_table_values(
...
+        case IntegralType(is_expression=True):
+            assert integral_type.codim in (0, 1)
+            assert integral_type.num_neighbours == 1
```
jorgensd:
```suggestion
        case IntegralType(is_expression=True):
            # FFCx tables for expression are generated as either interior cell points
            # or points on a facet
            assert integral_type.codim in (0, 1)
            assert integral_type.num_neighbours == 1
```

PR #790, opened 2025-10-13
Python 3.14 CI testing

schnellerhase:
- Extend test matrix to all {OS}x{Python} combinations
- CI support for Python 3.14
- Remove Python 3.9 configuration
- numba 0.63 is Python 3.14 compatible


jorgensd:
I guess this is good to go.
@jhale  what do you think?

Issue #687, opened 2024-04-26
Remove `ufcx.h`


garth-wells:
With the removal of finite elements and dofmaps from UFCx in #681, there isn't much left apart from the kernel interface. Maybe we can removed `ufcx.h` and simply document the interface?

jhale:
I'm currently trying to make ufcx.h and subsequent cffi compilation C11 compliant (complex number support is optional). So if it was removed that would simplify things a lot.

jhale:
Here is the branch to make ufcx.h C11 compatible: https://github.com/FEniCS/ffcx/blob/jhale/c11compliance/ffcx/codegeneration/ufcx.h

But then CFFI complains because it cannot deal with C preprocessor macros.

jhale:
Any remaining necessary metadata about a form/expression could be written to a flat file format rather than being packaged up in a c-struct. This would also have the advantage of being easier to use the metadata across languages - the consumer would only have to worry about a single C interface for the kernel calls.

PR #801, opened 2025-11-26
Add `numba` backend

schnellerhase:
Based on @chrisrichardson's work at `chris/languages-3`

Introduces an optional `numba` backend for `FFCx`. In particular the code passes through the same `C`-function interface of `ufcx.h`, and therefore the standard assemblers of `DOLFINx` are compatible.

> [!IMPORTANT]  
> This does not introduce a (non-optional) dependency on `numba` - code generation is independent of compile.

Tested functionally on a simple poisson problem, and generation across all demo files.

Generalises some naming schemes to separate language specific implementation from functional concept.


jhale:
Would it be possible for this to be a Python backend, and then allow the user to use the JIT/intepreter tool they prefer, e.g. pure Python, Numba, JAX?

ffcx/codegeneration/numba/expressions.py
```diff
@@ -0,0 +1,135 @@
...
+    tensor_size = 1
+    for dim in ir.expression.tensor_shape:
+        tensor_size *= dim
+    n_coeff = 1000
+    n_const = 1000
+    header = f"""
+    A = numba.carray(_A, ({tensor_size}))
+    w = numba.carray(_w, ({n_coeff}))
+    c = numba.carray(_c, ({n_const}))
+    coordinate_dofs = numba.carray(_coordinate_dofs, (1000))
```
chrisrichardson: We need to figure out the correct size for these arrays

chrisrichardson:
Possibly could output pure python, but numba also requires some extras, like `carray` to translate data. I guess this could be optional.

schnellerhase:
> Would it be possible for this to be a Python backend, and then allow the user to use the JIT/intepreter tool they prefer, e.g. pure Python, Numba, JAX?

A pure python version is very similar to the Numba one and should be easy to add. Maybe best to do in a second step though.

For JAX and/or numpy we are hitting the problem of the algebraic lowering in the IR - so translating the IR to a vectorised language will not result in what one would expect from a JAX/numpy kernel. Having a proper support for JAX or numpy I believe will require a redesign to maintain the tensor structure of operations in the IR.

ffcx/codegeneration/numba/expressions.py
schnellerhase: Fixed the integral version (since I have tests for that already) [here](https://github.com/FEniCS/ffcx/pull/801/files#diff-30e36f6eeb93ee55f633b257e92010b17752124af26923a11febcb8979f984eaR57-R73). Should we move those computations into `compute_*_ir` and just access here? Only for the shape of `n_entity_local_index` I am having a hard time to figure out the exact value, should this just be deduced from measure or do we have some functionality inplace to extract this?

ffcx/codegeneration/numba/expressions.py
```diff
@@ -0,0 +1,116 @@
...
+    eg = ExpressionGenerator(ir, backend)
+
+    d: dict[str, str | int] = {}
+    d["name_from_uflfile"] = ir.name_from_uflfile
+    d["factory_name"] = factory_name
+    parts = eg.generate()
+
+    # tabulate_expression
+    tensor_size = np.prod(ir.expression.shape)
+    tensor_size *= 3  # TODO: number of evaluation points - where to get?
```
schnellerhase: Don't know, from which IR member I can extract this info.

ffcx/codegeneration/numba/file_template.py
```diff
@@ -0,0 +1,47 @@
...
+# This code was generated with the following options:
+#
+{options}
+
+import numba
+import numpy as np
+import math
+
+# ufcx enums
+interval = 10
```
jorgensd, line 29: We should find a way to ensure that these are consistent cross implementation backend

schnellerhase: I think this relates to https://github.com/FEniCS/ffcx/issues/687: currently the `C` implementation includes it from `ufcx.h` and the `numba` side defines it customly in the `file_template.py`. The generic solution would be to maintain the values in `FFCx` and inject into the templates, for the different backends. Without getting rid of `ufcx.h`, we need to synchronise a `C`-header and a Python file, not sure how to do that nicely.

demo/MassAction.py
```diff
@@ -6,6 +6,7 @@
 """Mass action demo."""
 
 import basix
+import basix.ufl
```
jorgensd, line 8: Why is this now required?

ffcx/main.py
```diff
@@ -29,6 +29,7 @@
 )
 parser.add_argument("--version", action="version", version=f"%(prog)s (version {FFCX_VERSION})")
 parser.add_argument("-o", "--output-directory", type=str, default=".", help="output directory")
+parser.add_argument("-L", "--language", type=str, default="C", help="target language")
```
jorgensd: Add list of choices here?

demo/MassAction.py
schnellerhase: Execution of the file standalone requires it for the call of `basix.ufl.element` - I am not entirely sure why this did not cause an issue for the tested invocation through `ffcx ...`.

jorgensd: Then we should probably remove the first import as basix.ufl also imports basix

ffcx/main.py
```diff
@@ -29,6 +29,9 @@
 )
 parser.add_argument("--version", action="version", version=f"%(prog)s (version {FFCX_VERSION})")
 parser.add_argument("-o", "--output-directory", type=str, default=".", help="output directory")
+parser.add_argument(
```
michalhabera: Shouldn't it be defined in `options.py`? It is not a command-line specific option.

ffcx/codegeneration/numba/expressions.py
```diff
@@ -0,0 +1,106 @@
...
+    size_c = sum(
+        np.prod(constant.ufl_shape, dtype=int)
+        for constant in ir.expression.original_constant_offsets.keys()
+    )
+    size_coords = ir.expression.number_coordinate_dofs * 3
+    size_local_index = 2  # TODO: this is just an upper bound, harmful?
+    size_permutation = 2 if ir.expression.needs_facet_permutations else 0
+
+    header = f"""
+    A = numba.carray(_A, ({points.size}))
```
michalhabera: Is this correct? Expression allows evaluation at points, tensor valued nodes and with argument dofs (i.e. if the expression is a linear form), see signature in https://github.com/FEniCS/ffcx/blob/main/ffcx/codegeneration/ufcx.h#L157

ffcx/codegeneration/codegeneration.py
```diff
@@ -39,12 +36,30 @@ class CodeBlocks(typing.NamedTuple):
...
+def generate_code(
+    ir: DataIR, options: dict[str, int | float | npt.DTypeLike]
+) -> tuple[CodeBlocks, tuple[str, str]]:
     """Generate code blocks from intermediate representation."""
     logger.info(79 * "*")
     logger.info("Compiler stage 3: Generating code")
     logger.info(79 * "*")
 
+    lang = options.get("language", "C")
+    # try:
```
michalhabera: Leftover code?

ffcx/compiler.py
```diff
@@ -115,12 +115,12 @@ def compile_ufl_objects(
...
+    code, suffixes = generate_code(ir, options)
     _print_timing(3, time() - cpu_time)
 
     # Stage 4: format code
     cpu_time = time()
     code_h, code_c = format_code(code)
     _print_timing(4, time() - cpu_time)
 
-    return code_h, code_c
+    return code_h, code_c, suffixes
```
michalhabera: This is a public API breaking change. Would there be a less intrusive option?

test/poisson.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2004-2007 Anders Logg
+# Copyright (C) 2004-20025 Anders Logg and Paul T. Kühner
```
michalhabera:
```suggestion
# Copyright (C) 2004-2025 Anders Logg and Paul T. Kühner
```

test/poisson.py
```diff
@@ -31,19 +31,27 @@
     dx,
     grad,
     inner,
+    tr,
 )
 
-mesh = Mesh(basix.ufl.element("P", "triangle", 2, shape=(2,)))
+mesh = Mesh(basix.ufl.element("P", "triangle", 1, shape=(2,)))
```
michalhabera, line 37: Why order 1 now?

test/poisson.py
```diff
@@ -31,19 +31,27 @@
...
 u = TrialFunction(space)
 v = TestFunction(space)
 f = Coefficient(space)
 
-kappa1 = Constant(mesh, shape=(2, 2))
-kappa2 = Constant(mesh, shape=(2, 2))
+kappa = Constant(mesh, shape=(2, 2))
 
-a = inner(kappa1, kappa2) * inner(grad(u), grad(v)) * dx
+a = tr(kappa) * inner(grad(u), grad(v)) * dx
```
michalhabera, line 49: Also this is a different math now. Any reason?

michalhabera (changes requested):
I like these efforts from @chrisrichardson and @schnellerhase . However, in mid-term, we should define a common interface and allow for custom user-defined language backends. For now, up to my few comments this looks mergeable, given that all code lives in a separate namespace and is independent.

test/poisson.py
schnellerhase: Both changes to the math are simplifications for the assembly tests in `test_numba.py`. Before this file was solely used for the command line interface test, which did not evaluate any math, so we do not reduce test coverage.

ffcx/codegeneration/codegeneration.py
schnellerhase: For the custom language support, will move to an issue.

ffcx/compiler.py
schnellerhase: Intrusive wise this is the best we can do I think.

However, this is one of the points we should iterate later for an improved interface for more backends - the assumption on header and implementation split is a `C` perspective: the output file structure is language dependant and should be handled there.

ffcx/codegeneration/numba/expressions.py
schnellerhase: Indeed a very wrong concept here. Changed to (tensor shape computed [here](https://github.com/FEniCS/ffcx/blob/818267a4f538d0eeab1afb54274b2580d2dbbaa0/ffcx/ir/representation.py#L530-L542))
```python
num_points = points.shape[0]
num_components = np.prod(ir.expression.shape, dtype=np.int32)
num_argument_dofs = np.prod(ir.expression.tensor_shape, dtype=np.int32)
size_A = num_points * num_components * num_argument_dofs
```

Updated test to assemble an expression of shape `(3, 2)` - previously breaking.

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,257 @@
...
+        """Format an array declaration."""
+        if arr.symbol.dtype == L.DataType.SCALAR:
+            dtype = "A.dtype"
+        elif arr.symbol.dtype == L.DataType.REAL:
+            dtype = "coordinate_dofs.dtype"
+        elif arr.symbol.dtype == L.DataType.INT:
+            dtype = "np.int32"
+        symbol = self.format(arr.symbol)
+        if arr.values is None:
+            return f"{symbol} = np.empty({arr.sizes}, dtype={dtype})\n"
```
michalhabera: Is the explicit `dtype` needed here?

schnellerhase: Yes they are, dropping them here always results in casts to `np.float64`. However produced code has a inferred type, so it still remains general:

```python
temp_0 = np.full((3,), 0, dtype=A.dtype)
```

michalhabera: Hm, what I don't like here is the deduction of dtype from either `A` or `coordinate_dofs`.
Why don't you always take the `A.dtype`?

schnellerhase: These data types are not necessarily the same. `A.dtype` is the tensor type, can be real or complex, while `coordinate_dofs.dtype` is always the corresponding real type, as it holds geometric information. Its more apparent in the C++ backend, where the template signature of the kernel lives in both of these types:

```c++
template <typename T, typename U>
void tabulate_tensor(T* A,
                     const T* RESTRICT w,
                     const T* RESTRICT c,
                     const U* RESTRICT coordinate_dofs,
                     const std::int32_t* RESTRICT entity_local_index,
                     const std::uint8_t* RESTRICT quadrature_permutation);
```

michalhabera: I know these two are not the same, but I am still confused. The `CFormatter` has `scalar_type` and `real_type`, both numpy dtypes ready to be used here. Why don't you use these? We need to avoid saying `coordinate_dofs` is the prototype for real_type, that will break more refined dtype logic very early.

schnellerhase: Right, changed to the `_dtype_to_name` type extraction as in the `CFormatter`.

ffcx/codegeneration/C/integrals.py
```diff
@@ -10,18 +10,21 @@
...
 from ffcx.codegeneration.utils import dtype_to_c_type, dtype_to_scalar_dtype
 from ffcx.ir.representation import IntegralIR
 
 logger = logging.getLogger("ffcx")
 
 
-def generator(ir: IntegralIR, domain: basix.CellType, options):
+def generator(
+    ir: IntegralIR, domain: basix.CellType, options: dict[str, int | float | npt.DTypeLike]
+):
```
jorgensd: Add return-type here?
Maybe we could expand the doc-string a bit, i.e. mention that the generator returns a declaration and implementation of the C code.

ffcx/codegeneration/numba/expressions.py
```diff
@@ -0,0 +1,110 @@
...
+from ffcx.codegeneration.expression_generator import ExpressionGenerator
+from ffcx.codegeneration.numba import expressions_template
+from ffcx.codegeneration.numba.implementation import Formatter
+from ffcx.ir.representation import ExpressionIR
+
+logger = logging.getLogger("ffcx")
+
+
+def generator(ir: ExpressionIR, options: dict[str, int | float | npt.DTypeLike]) -> tuple[str, str]:
+    """Generate UFC code for an expression."""
```
jorgensd, line 24: Here we have the type-hint!:)

But maybe add some more descriptions to the docstring, ref the one above.

ffcx/codegeneration/numba/expressions.py
```diff
@@ -0,0 +1,110 @@
...
+    num_argument_dofs = np.prod(ir.expression.tensor_shape, dtype=np.int32)
+    size_A = num_points * num_components * num_argument_dofs  # ref. ufcx.h
+    size_w = sum(coeff.ufl_element().dim for coeff in ir.expression.coefficient_offsets.keys())
+    size_c = sum(
+        np.prod(constant.ufl_shape, dtype=int)
+        for constant in ir.expression.original_constant_offsets.keys()
+    )
+    size_coords = ir.expression.number_coordinate_dofs * 3
+    size_local_index = 2  # TODO: this is just an upper bound, harmful?
+    size_permutation = 2 if ir.expression.needs_facet_permutations else 0
```
jorgensd, line 57: Could you comment on why these are two, and not 1?
As far as I can tell, currently we only pass in a single local index or single permutation to the kernels, thus the re-cast would only be required if we had more data per cell.

Thinking about it, maybe using the local-index with more data could be used to pass the cell-index into the kernel (needed for some custom stuff). Currently cell integrals has a null-ptr as input, and we could take that into account here, no?

ffcx/codegeneration/numba/file.py
```diff
@@ -0,0 +1,47 @@
...
+from ffcx.codegeneration import __version__ as UFC_VERSION
+from ffcx.codegeneration.numba import file_template
+
+logger = logging.getLogger("ffcx")
+
+
+def generator(
+    options: dict[str, int | float | npt.DTypeLike],
+) -> tuple[tuple[str, str], tuple[str, str]]:
+    """Generate UFC code for file output."""
```
jorgensd: You know I like docstrings!:)

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+"""Numba implementation for output."""
+
+import numpy as np
+from numpy import typing as npt
+
+import ffcx.codegeneration.lnodes as L
+from ffcx.codegeneration.utils import dtype_to_scalar_dtype
+
+
+def build_initializer_lists(values):
```
jorgensd:
```suggestion
def build_initializer_lists(values: np.ndarray) -> str:
```
Should we add an option for the string formatting of the inner values, so that we could do single precision formatting if that is wanted, instead of printing double precision, but later casting dtype as single?

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+
+    scalar_type: np.dtype
+    real_type: np.dtype
+
+    def __init__(self, dtype: npt.DTypeLike) -> None:
+        """Initialise."""
+        self.scalar_type = np.dtype(dtype)
+        self.real_type = dtype_to_scalar_dtype(dtype)
+
+    def _dtype_to_name(self, dtype) -> str:
```
jorgensd:
```suggestion
    def _dtype_to_name(self, dtype: L.DataType) -> str:
```

Note that when we finally get to Python 3.11, we can make this simpler: https://docs.python.org/3/library/enum.html#enum.StrEnum
We could also consider using match-case: https://docs.python.org/3/tutorial/controlflow.html#match-statements

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+        if dtype == L.DataType.INT:
+            return f"np.{np.int32}"
+        if dtype == L.DataType.BOOL:
+            return f"np.{np.bool}"
+        raise ValueError(f"Invalid dtype: {dtype}")
+
+    def format_section(self, section: L.Section) -> str:
+        """Format a section."""
+        # add new line before section
+        comments = "# ------------------------ \n"
```
jorgensd: Should we use `format_comment` on each of these?

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+    def format_statement_list(self, slist: L.StatementList) -> str:
+        """Format a list of statements."""
+        output = ""
+        for s in slist.statements:
+            output += self.format(s)
+        return output
+
+    def format_comment(self, c: L.Comment) -> str:
+        """Format a comment."""
+        return "# " + c.comment + "\n"
```
jorgensd:
```suggestion
        return f"# {c.comment} \n"
```

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+        dtype = arr.symbol.dtype
+        typename = self._dtype_to_name(dtype)
+
+        symbol = self.format(arr.symbol)
+        if arr.values is None:
+            return f"{symbol} = np.empty({arr.sizes}, dtype={typename})\n"
+        elif arr.values.size == 1:
+            return f"{symbol} = np.full({arr.sizes}, {arr.values[0]}, dtype={typename})\n"
+        av = build_initializer_lists(arr.values)
+        av = "np.array(" + av + f", dtype={typename})"
```
jorgensd: Why not use an f-string for the full string here?

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def format_nary_op(self, oper: L.NaryOp) -> str:
+        """Format a n argument operation."""
+        # Format children
+        args = [self.format(arg) for arg in oper.args]
+
+        # Apply parentheses
+        for i in range(len(args)):
+            if oper.args[i].precedence >= oper.precedence:
+                args[i] = "(" + args[i] + ")"
```
jorgensd:
```suggestion
                args[i] = f"({args[i]})"
```

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def format_for_range(self, r: L.ForRange) -> str:
+        """Format a loop over a range."""
+        begin = self.format(r.begin)
+        end = self.format(r.end)
+        index = self.format(r.index)
+        output = f"for {index} in range({begin}, {end}):\n"
+        b = self.format(r.body).split("\n")
+        for line in b:
+            output += f"    {line}\n"
```
jorgensd, line 183: Should we make the tabulation here a configuration parameter?

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+    def format_conditional(self, s: L.Conditional) -> str:
+        """Format a conditional."""
+        # Format children
+        c = self.format(s.condition)
+        t = self.format(s.true)
+        f = self.format(s.false)
+
+        # Apply parentheses
+        if s.condition.precedence >= s.precedence:
+            c = "(" + c + ")"
```
jorgensd:
```suggestion
            c = f"({c})"
```

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+        if s.true.precedence >= s.precedence:
+            t = "(" + t + ")"
+        if s.false.precedence >= s.precedence:
+            f = "(" + f + ")"
```
jorgensd:
```suggestion
        if s.true.precedence >= s.precedence:
            t = f"({t})"
        if s.false.precedence >= s.precedence:
            f = f"({f})"
```

ffcx/codegeneration/numba/implementation.py
```diff
@@ -0,0 +1,274 @@
...
+            "atan": "arctan",
+            "atan2": "arctan2",
+            "acosh": "arccosh",
+            "asinh": "arcsinh",
+            "atanh": "arctanh",
+        }
+        function = function_map.get(f.function, f.function)
+        args = [self.format(arg) for arg in f.args]
+        if "bessel" in function:
+            return "0"
```
jorgensd: Is this correct? Should it throw an error if we don't have bessel, or should we default to `scipy special`: https://github.com/numba/numba/issues/2312

ffcx/compiler.py
```diff
@@ -89,7 +89,7 @@ def compile_ufl_objects(
     object_names: dict[int, str] | None = None,
     prefix: str | None = None,
     visualise: bool = False,
-) -> tuple[str, str]:
+) -> tuple[str, str, tuple[str, str]]:
```
jorgensd: Document out-type.

ffcx/compiler.py
michalhabera: Sure, we'll improve later, but please make sure this MR does not break API for ffcx compiler callers. What do you need suffixes for? In my understanding, any caller of `compile_ufl_objects` knows the language backend he asked for, so should not depend on the suffix returned.

ffcx/codegeneration/numba/implementation.py
schnellerhase: Will raise for now, as this requires an additional import to handle properly. But add the link so we can add the script calls in the future.

ffcx/codegeneration/numba/implementation.py
schnellerhase: Is there a benefit associated with doing the cast here, instead of on type declaration?

ffcx/codegeneration/numba/expressions.py
schnellerhase: From `ufcx.h`:
>  For integrals not on interior facets, this argument has no effect and a null pointer can be passed. For interior facets the array will have size 2 (one permutation for each cell adjacent to the facet).

The two case always is associated with the interior facet case.

schnellerhase: > Thinking about it, maybe using the local-index with more data could be used to pass the cell-index into the kernel (needed for some custom stuff). Currently cell integrals has a null-ptr as input, and we could take that into account here, no?

I believe this is what @sclaus2 is trying to do in https://github.com/FEniCS/ffcx/pull/804 and https://github.com/FEniCS/dolfinx/pull/4013 and should be decided there. Not sure I like the approach, since it is altering the interpretation of the interface depending on integral type. That's bug prone. Also, how does this scale to interior facets, where one needs the `local_index` parameter - by packing both cell index and local facet index?

sclaus2: Exactly right. For runtime quadrature rules which in general change in each cell I need to know at least the `loop_index` inside integral kernels. I will open a PR on several ways on how this could be passed to the kernel for custom_data (Option 1:  Wrap custom_data in std::function of `loop_index`, Option 2: Build struct that contains local_index and custom_data, local_index is then assigned in assembly loop , Option 3: Append it to entity_local_index). In the case of custom_data all other indices can then be inferred from the `loop_index`, by for example including the entities vector in custom_data.

demo/test_demos.py
```diff
@@ -1,60 +1,87 @@
...
+    f
+    for f in demo_dir.iterdir()
+    if f.suffix == ".py" and not f.stem.endswith("_numba") and f != Path(__file__)
+]
+
+skip_complex = ["BiharmonicHHJ", "BiharmonicRegge", "StabilisedStokes"]
+
+
+def skip_unsupported(test):
+    """Dcecorate test case to skip unsupported cases."""
```
michalhabera:
```suggestion
    """Decorate test case to skip unsupported cases."""
```

ffcx/codegeneration/C/integrals.py
```diff
@@ -10,19 +10,32 @@
...
-def generator(ir: IntegralIR, domain: basix.CellType, options):
-    """Generate C code for an integral."""
+def generator(
+    ir: IntegralIR, domain: basix.CellType, options: dict[str, int | float | npt.DTypeLike]
+) -> tuple[str, str]:
+    """Generate C code for an integral.
+
+    Args:
+        ir: IR of the integral
+        domain: BASIx cell type
```
michalhabera:
```suggestion
        domain: basix cell type
```

ffcx/codegeneration/numba/integrals.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def generator(
+    ir: IntegralIR, domain: basix.CellType, options: dict[str, int | float | npt.DTypeLike]
+) -> tuple[str, str]:
+    """Generate numba code for an integral.
+
+    Args:
+        ir: IR of the integral
+        domain: BASIx cell type
```
michalhabera:
```suggestion
        domain: basix cell type
```

ffcx/codegeneration/numba/integrals.py
```diff
@@ -0,0 +1,110 @@
...
+
+    """
+    logger.info("Generating code for integral:")
+    logger.info(f"--- type: {ir.expression.integral_type}")
+    logger.info(f"--- name: {ir.expression.name}")
+
+    # Note: In contrast to the C implementation the actual code is type agnostic, thus not part of
+    #       naming. This is only true for the numba code prior to compilation - the JITed versions
+    #       are different.
+    factory_name = ir.expression.name
```
michalhabera: C integral generator has
```python
factory_name = f"{ir.expression.name}_{domain.name}"
```
What do you mean with "not part of naming"?

schnellerhase: Oh good catch. Thought this is where we add the type specific suffix `_float64` to the names. But that happens at another step. This should be the same as in the C code.

jorgensd:
What if we rather make this a `jitclass`, which initializes the tables once: https://numba.readthedocs.io/en/stable/user/jitclass.html?highlight=jitclass ?

michalhabera (approved):
Happy to have this merged as an exploratory language backend.

chrisrichardson (approved):

PR #803, opened 2025-11-30
Add `C++` backend

schnellerhase:
Based on @chrisrichardson's work at `chris/languages-3`.

The `C++` backend, in particular in comparison to the `C` backend, allows for type injection into the kernels, due to `C++`'s template support. This enables arbitrary precision kernels as well as user defined non-fundamental types.

TODO:
- [ ] expression support
- [ ] unit test of functionality
- [x] add `__restrict__`? - figure out how to automatically remove if compiler does not support
- [ ] signature change of the kernel, do we want to be compliant with `C` or leverage also here the upsides of `C++`? `type* -> std::mdspan<tpye, N>`, `void* custom_data -> template type`.


garth-wells:
> @schnellerhase - can you add some more context to the PR, why is a c++ backend desirable?

C++ backend will be very nice if the kernel can be passed as a lambda function to the assembler, which allows inlining. 

Using constexpr will be handy too. And we won’t have to hardwire the float type.

Issue #805, opened 2025-12-07
Support user defined language backends


schnellerhase:
With https://github.com/FEniCS/ffcx/pull/801 we introduce a second language backend. This should be extended to support user definitions of language backends.

User provides a python module `my-language` (equivalent of the `C` and `numba` module insides `FFCx`), and can control its usage through the command line options, with

```shell
ffcx --language my-module.my-lang Poisson.py
```

There should be some work put into generalising the interfaces and enhance reusability of the currently present routines to allow for easy (external) extension.

PR #806, opened 2025-12-10
Backend formatter interface

schnellerhase:
Switches to usage of `singledispatchmethod` for simplified (non custom string based lookup) type dispatching inside formatters and renames `implementation.py -> formatter.py`

Locks down interface to general signature (rest is implementation detail):

```python
class FormatterImpl:
```

Which should be called as

```python
format = FormatterImpl(opts)
source = format(ir)
```

Work towards https://github.com/FEniCS/ffcx/issues/805.

### DIFF
--- a/test/test_numba.py
+++ b/test/test_numba.py
@@ -8,13 +8,14 @@ import ctypes
 import importlib
 from pathlib import Path
 
-import numba
 import numpy as np
 import numpy.typing as npt
 import pytest
 
 import ffcx.main
 from ffcx.codegeneration.utils import dtype_to_scalar_dtype, numba_ufcx_kernel_signature
+
+numba = pytest.importorskip("numba")
 
 
 def wrap_kernel(scalar_type, real_type):


ffcx/codegeneration/C/implementation.py
```diff
@@ -184,11 +185,27 @@ def _build_initializer_lists(self, values):
...
+    def _(self, slist: L.StatementList) -> str:
         """Format a statement list."""
-        return "".join(self.format(s) for s in slist.statements)
+        return "".join(self(s) for s in slist.statements)
 
-    def format_section(self, section) -> str:
+    @_format.register
+    def _(self, slist: L.StatementList) -> str:
+        """Format a statement list."""
+        return "".join(self(s) for s in slist.statements)
```
jorgensd, line 197:
```suggestion
      @singledispatchmethod
    def __call__(self, obj) -> str:
        """Format an L Node."""
        raise NotImplementedError(f"Can not format object to type {type(obj)}")


    @__call__.register
    def _(self, slist: L.StatementList) -> str:
        """Format a statement list."""
        return "".join(self(s) for s in slist.statements)
```
Could we rather do singledispatch on `__call__`, as it is slightly less code.

Also remove the duplicate overload of statementlist.

ffcx/codegeneration/C/implementation.py
```diff
@@ -269,52 +292,73 @@ def format_binary_op(self, oper) -> str:
...
         return f"{oper.op}{arg}"
 
-    def format_literal_float(self, val) -> str:
+    @_format.register
+    def _(self, val: L.Neg) -> str:
+        return self._format_unary_op(val)
+
+    @_format.register
+    def _(self, val: L.Not) -> str:
+        return self._format_unary_op(val)
```
jorgensd:
```suggestion
    @__call__.register(L.Neg)
    @__call__.register(L.Not)
    def _(self, oper: L.Not | L.Neg) -> str:
        """Format a unary operation."""
        arg = self(oper.arg)
        if oper.arg.precedence >= oper.precedence:
            return f"{oper.op}({arg})"
        return f"{oper.op}{arg}"
```
which can be further simplified with Python 3.11 as min level:
https://github.com/FEniCS/ufl/pull/365#discussion_r2055613546

As a side-note, I check, and we do not have any tests that check the `L.Not` in FFCx.

ffcx/codegeneration/C/implementation.py
```diff
@@ -269,52 +292,73 @@ def format_binary_op(self, oper) -> str:
...
         return f"{lhs} {expr.op} {rhs};\n"
 
-    def format_conditional(self, s) -> str:
+    @_format.register
+    def _(self, expr: L.Assign) -> str:
+        return self._format_assign(expr)
+
+    @_format.register
+    def _(self, expr: L.AssignAdd) -> str:
+        return self._format_assign(expr)
```
jorgensd:
```suggestion
    @__call__.register(L.Assign)
    @__call__.register(L.AssignAdd)    
    def _(self, expr: L.Assign|L.AssignAdd) -> str:
        """Format an assignment."""
        rhs = self(expr.rhs)
        lhs = self(expr.lhs)
        return f"{lhs} {expr.op} {rhs};\n"
```

ffcx/codegeneration/numba/implementation.py
```diff
@@ -46,7 +48,17 @@ def _dtype_to_name(self, dtype: L.DataType) -> str:
             return f"np.{np.bool}"
         raise ValueError(f"Invalid dtype: {dtype}")
 
-    def format_section(self, section: L.Section) -> str:
+    def __call__(self, obj) -> str:
```
jorgensd: Same comments here.

ffcx/codegeneration/numba/implementation.py
```diff
@@ -137,21 +157,23 @@ def format_binary_op(self, oper: L.BinOp) -> str:
...
-    def format_not(self, val: L.Not) -> str:
+    @_format.register
+    def _(self, val: L.Not) -> str:
         """Format not operation."""
-        arg = self.format(val.arg)
+        arg = self(val.arg)
         return f"not({arg})"
 
-    def format_andor(self, oper: L.And | L.Or) -> str:
+    def _format_and_or(self, oper: L.And | L.Or) -> str:
```
jorgensd: Same comment as for the C implementation regarding using multiple dispatch on unions.

ffcx/codegeneration/numba/implementation.py
```diff
@@ -164,41 +186,64 @@ def format_andor(self, oper: L.And | L.Or) -> str:
...
         return output
 
-    def format_statement(self, s: L.Statement) -> str:
+    @_format.register
+    def _(self, s: L.Statement) -> str:
         """Format a statement."""
-        return self.format(s.expr)
+        return self(s.expr)
+
+    def _format_assign(self, expr) -> str:
```
jorgensd: Same

ffcx/codegeneration/C/implementation.py
schnellerhase: Tried, but I think `__call__` is not supported by `singledispatch`.

jorgensd: I tried it locally and it seemed to work for me:P

schnellerhase: I stand corrected.

ffcx/codegeneration/C/implementation.py
schnellerhase: Nice!

ffcx/codegeneration/C/formatter.py
```diff
@@ -197,28 +206,30 @@ def format_section(self, section) -> str:
...
             body = body[:-2] + "}\n"
 
         body += "// ------------------------ \n"
         return str(comments + declarations + body)
 
-    def format_comment(self, c: L.Comment) -> str:
+    @__call__.register
+    def _(self, c: L.Comment) -> str:
         """Format a comment."""
         return "// " + c.comment + "\n"
```
jorgensd:
```suggestion
        return f"// {c.comment}"\n"
```

ffcx/codegeneration/C/formatter.py
```diff
@@ -269,52 +284,61 @@ def format_binary_op(self, oper) -> str:
         # Return combined string
         return f"{lhs} {oper.op} {rhs}"
 
-    def format_unary_op(self, oper) -> str:
+    @__call__.register(L.Neg)
+    @__call__.register(L.Not)
+    def _(self, oper) -> str:
```
jorgensd:
```suggestion
    def _(self, oper: L.Neg | L.Not) -> str:
```

ffcx/codegeneration/numba/formatter.py
```diff
@@ -55,36 +64,39 @@ def format_section(self, section: L.Section) -> str:
...
         return comments + declarations + body
 
-    def format_statement_list(self, slist: L.StatementList) -> str:
+    @__call__.register
+    def _(self, slist: L.StatementList) -> str:
         """Format a list of statements."""
         output = ""
         for s in slist.statements:
-            output += self.format(s)
+            output += self(s)
```
jorgensd: Is there a reason for this one not looking like the C formatter version of statement list (With a quick glance I can't spot the difference with 
```python
 return "".join(self(s) for s in slist.statements)
```

PR #808, opened 2025-12-12
Centralise common backend routines and continue backend interface

schnellerhase:
Centralises common backend logic for reusability in different backends.
- `template_keys`: extracts template keys, used for checking all keys matched by input data
- `integral_data`: computes unrolled sorted integral data of a `FormIR`
- `tensor_sizes`: computes the sizes of the tensors associated with the kernel invocation of a form or an expression

Consistently names the implementation files in singular (`expressions.py` -> `expression.py`, etc.)

All includes in code generation now pass though the module level, i.e. `__init__.py` of `C/` or `numba/` - no longer enforcing a given file structure to be present in the backend (that's an implementation detail). This continues the work to define the language backend interface.

```python
"""Generation of MyBackend code."""

class FormatterImpl:

class expression: # (or namespace)
    def generator(ir: ExpressionIR, options: dict[str, int | float | npt.DTypeLike]) -> tuple[str, str]: ...

class integral: # (or namespace)
    def generator(ir: IntegralIR, domain: basix.CellType, options: dict[str, int | float | npt.DTypeLike]) -> tuple[str, str]: ...

class form: # (or namespace)
     def generator(ir: FormIR, options: dict[str, int | float | npt.DTypeLike]) -> tuple[str, str]: ...

class file: # (or namespace)
    def generator(options: dict[str, int | float | npt.DTypeLike]) -> tuple[tuple[str, str], tuple[str, str]]: ...
```

The only thing that needs to be addressed/improved on after this for https://github.com/FEniCS/ffcx/issues/805 is the handling of output files, the split into declaration and implementation (or header and source) should be generalised away from the backend.

### DIFF
--- a/ffcx/codegeneration/numba/__init__.py
+++ b/ffcx/codegeneration/numba/__init__.py
@@ -1,6 +1,6 @@
 """Generation of numba code."""
 
-from ffcx.codegeneration.C import expression, file, form, integral
+from ffcx.codegeneration.numba import expression, file, form, integral
 
 from .formatter import Formatter
 


ffcx/codegeneration/common.py
```diff
@@ -0,0 +1,100 @@
...
+    """Sorted integral data."""
+
+    names: list[str]
+    ids: list[int]
+    offsets: list[int]
+    domains: list[list[basix.CellType]]
+
+
+def integral_data(ir: FormIR) -> IntegralData:
+    """Extracts sorted intergral data from form."""
```
michalhabera:
```suggestion
    """Extracts sorted integral data from form."""
```

michalhabera (approved):

PR #813, opened 2025-12-17
Move file separation logic to backend

schnellerhase:
Finalises the general backend interface and adds injection of external modules through `--language mymodule.mylanguage`.
Backend interface is

```python

class FormatterImpl:

class expression: # (or namespace)
     def generator(ir: ExpressionIR, options: dict[str, int | float | npt.DTypeLike]) -> tuple[str, ...]: ...

class integral: # (or namespace)
     def generator(ir: IntegralIR, domain: basix.CellType, options: dict[str, int | float | npt.DTypeLike]) -> tuple[str, ...]: ...

class form: # (or namespace)
    def generator(ir: FormIR, options: dict[str, int | float | npt.DTypeLike]) -> tuple[str, ...]: ...

class file: # (or namespace)
    suffixes: tuple[str, ...]
    def generator(options: dict[str, int | float | npt.DTypeLike]) -> tuple[tuple[str,...], tuple[str, ...]]: ...
```

In particular, in comparison to https://github.com/FEniCS/ffcx/pull/808, note the variadic sizes of the returned tuples and the new `file.suffixes` variable. Backends are free to produce an arbitrary number of output files now.

Fixes https://github.com/FEniCS/ffcx/issues/805.

### DIFF
--- a/ffcx/formatting.py
+++ b/ffcx/formatting.py
@@ -32,15 +32,14 @@ def format_code(code_blocks: CodeBlocks) -> list[str]:
     code = [""] * len(code_blocks[0][0])
 
     for block in code_blocks:
-        for i, c in enumerate(code):
-            c += "".join([b[i] for b in block])
+        for i in range(len(code)):
+            code[i] += "".join([c[i] for c in block])
 
     return code
 
 
 def write_code(code: list[str], prefix: str, suffixes: tuple[str, ...], output_dir: str) -> None:
     """Write code to files."""
-outfile = Path(output_dir) / prefix
-for source, suffix in zip(code, suffixes, strict=True):
-       with open(outfile.with_suffix(suffix), "w") as file:
+    for source, suffix in zip(code, suffixes, strict=True):
+        with open(Path(output_dir) / (prefix + suffix), "w") as file:
             file.write(source)


ffcx/compiler.py
jorgensd: Maybe we could expand a bit on the documentation here now?:)

ffcx/formatting.py
```diff
@@ -29,40 +29,17 @@ def format_code(code: CodeBlocks) -> tuple[str, str]:
...
-    code_c = ""
-    code_h = ""
-    for parts_code in code:
-        code_h += "".join([c[0] for c in parts_code])
-        code_c += "".join([c[1] for c in parts_code])
+    code = [""] * len(code_blocks[0][0])
 
-    return code_h, code_c
+    for block in code_blocks:
+        for i in range(len(code)):
```
jorgensd, line 35: This could be an enumerate, i.e.
```python
for block in code_blocks:
    for i, c in enumerate(code):
       c += "".join([b[i] for b in block])
```
?

ffcx/formatting.py
```diff
@@ -29,40 +29,17 @@ def format_code(code: CodeBlocks) -> tuple[str, str]:
...
-
-    if suffixes[0] is not None:
-        _write_file(code_h, filename_stem, suffixes[0], output_dir)
-    if suffixes[1] is not None:
-        _write_file(code_c, filename_stem, suffixes[1], output_dir)
+def write_code(code: list[str], prefix: str, suffixes: tuple[str, ...], output_dir: str) -> None:
+    """Write code to files."""
+    for source, suffix in zip(code, suffixes):
+        with open(Path(output_dir) / (prefix + suffix), "w") as file:
+            file.write(source)
```
jorgensd: What about something like
```suggestion
outfile = Path(output_dir) / prefix
for source, suffix in zip(code, suffixes, strict=True):
       with open(outfile.with_suffix(suffix), "w") as file:
            file.write(source)
```

ffcx/formatting.py
schnellerhase: Weirdly this breaks - no clue why, maybe the named tuple and enumerate don't play nicely? - reverted.

PR #817, opened 2026-01-07
Fix duplicate CI runs

schnellerhase:
PRs currently run duplicate CI runs, once triggered through the `push` and once through the `pull_request` event.

- removes on push trigger for every branch but `main`.
- add pull request trigger for any branch (stacking PRs)


schnellerhase:
The amount of redundant work we perform is quite massive atm. The real fix would be to run either pull request or on push but that is not possible with GitHub actions https://github.com/orgs/community/discussions/26276. 

On a branch one can always comment out the single line https://github.com/FEniCS/ffcx/pull/817/changes#diff-cbf97851bdfebccf2fcdd8848c6a93adb8d8affd5c6b1faf00238d528c3ef6cbR10 to reactivate on push CIs or trigger through a workflow. Both really not nice options, but I would say preferable to running both on every PR push. In DOLFINx we do the same to avoid duplicate runs.

PR #818, opened 2026-02-09
Define language backend interface as protocol

schnellerhase:
Follows @finsberg and @jorgensd approach used in https://github.com/jorgensd/adios4dolfinx/pull/193.

### DIFF
--- a/ffcx/codegeneration/interface.py
+++ b/ffcx/codegeneration/interface.py
@@ -31,17 +31,12 @@ class Formatter(Protocol):
         ...
 
 
-def file_generator(options: dict[str, int | float | npt.DTypeLike]) -> tuple[tuple[str], ...]:
-    """File to source string.
+"""File to source string.
 
-    Arguments:
-        options: Options to pass to the generator backend through `ffcx_options`.
-
-    Note:
-        Needs to be callable as file.generator.
-    """
-    ...
-
+Note:
+    Needs to be callable as file.generator.
+"""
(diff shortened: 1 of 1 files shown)


ffcx/codegeneration/interface.py
```diff
@@ -4,22 +4,62 @@
...
-        raise NotImplementedError
+        ...
+
+
+"""File to source string.
+
+Note:
+    Needs to be callable as file.generator.
+"""
+file_generator = Callable[[dict[str, int | float | npt.DTypeLike]], tuple[tuple[str], ...]]
```
jorgensd: Would it be better to define this as:
```python
def file_generator(options: dict[str, int|float|npt.DTypeLike) -> tuple[tuple[str], ...]]:
    """File to source string.
    Arguments:
        options: Options to pass to the generator backend through `ffcx_options`.

    Note:
        Needs to be callable as file.generator.
"""

jorgensd: This would then hold for the subsequent functions as well.

schnellerhase: Will check if this works now - I had it like this before and it caused some problems. But had to alter different parts as well, so maybe we can return to this.

schnellerhase: Doesn't work. The `_file: interface.file_generator = file.generator` mypy checks seem to only pass for explicit Callable's not for stubs.

jorgensd: Ok.

PR #828, opened 2026-04-01
Allow Expressions with submeshes

jorgensd:
With submesh support, it is quite common that one want to evaluate an expression that uses quantities from both the parent and submesh.

This PR extends expression to be able to support:
- Submesh of codim 0
- Submesh of codim 1 (one sided facet expressions)

### DIFF
--- a/ffcx/ir/representation.py
+++ b/ffcx/ir/representation.py
@@ -606,6 +606,7 @@ def _compute_expression_ir(
         cell = None
     elif len(domains) == 1:
         expr_domain = domains[0]
+        cell = expr_domain.ufl_cell()
     else:
         expr_domain = max(domains, key=lambda domain: domain.topological_dimension)
         cell = expr_domain.ufl_cell()
--- a/ffcx/ir/representation.py
+++ b/ffcx/ir/representation.py
@@ -596,8 +596,6 @@ def _compute_expression_ir(
     points = expr[1]
     expr = expr[0]
 
- 
-
     domains = ufl.domain.extract_domains(expr)
     if len(domains) == 0:
         # This case corresponds to a spatially constant expression
--- a/ffcx/ir/representation.py
+++ b/ffcx/ir/representation.py
@@ -596,18 +596,8 @@ def _compute_expression_ir(
     points = expr[1]
     expr = expr[0]
 
-    domains = ufl.domain.extract_domains(expr)
-    if len(domains) == 0:
-        # This case corresponds to a spatially constant expression
-        # without any dependencies
-        expr_domain = None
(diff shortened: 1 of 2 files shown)


ffcx/ir/representation.py
```diff
@@ -596,12 +596,18 @@ def _compute_expression_ir(
...
         # This case corresponds to a spatially constant expression
         # without any dependencies
+        expr_domain = None
         cell = None
+    elif len(domains) == 1:
+        expr_domain = domains[0]
+        cell = expr_domain.ufl_cell()
+    else:
+        expr_domain = max(domains, key=lambda domain: domain.topological_dimension)
+        cell = expr_domain.ufl_cell()
```
schnellerhase:
```suggestion
    expr_domain = max(ufl.domain.extract_domains(expr), default=None, key=lambda d: d.topological_dimension)
    cell = expr_domain.ufl_cell() if expr_domain else None
```

schnellerhase (commented):
LGTM, can we add a simple unit test to check the new code path?

test/test_jit_expression.py
```diff
@@ -482,3 +482,76 @@ def test_coordinate_free_expression(compile_args):
...
+    edge_normals = np.cross(edges, face_normal)[:, :2]
+
+    # 4. Normalize the edge normals
+    norms = np.linalg.norm(edge_normals, axis=1, keepdims=True)
+    edge_normals_normalized = edge_normals / norms
+
+    u_coeffs = np.array([0.1, 0.5], dtype=dtype)
+    consts = np.array([], dtype=dtype)
+    entity_index = np.array([0], dtype=np.intc)
+    quad_perm = np.array([0], dtype=np.dtype("uint8"))
```
schnellerhase:
```suggestion
    quad_perm = np.array([0], dtype=np.uint8)
```

test/test_jit_expression.py
```diff
@@ -482,3 +482,76 @@ def test_coordinate_free_expression(compile_args):
...
+    output = np.zeros(points.shape[0] * 2, dtype=dtype)
+
+    # Define mesh
+    coords = np.array([[0.1, 0.3, 0], [2, 0, 0.0], [0, 1, 0.0]], dtype=dtype)
+
+    # Define exact normals
+    v1 = coords[1] - coords[0]
+    v2 = coords[2] - coords[0]
+    face_normal = np.cross(v1, v2)
+    face_normal /= np.linalg.norm(face_normal)
```
schnellerhase:
```suggestion
    face_normal = np.array([0, 0, 1], dtype=dtype)
```

PR #672, opened 2024-02-21
Expressions on facets

jorgensd:
Makes it possible to use expression tabulation on facets.

As for exterior facet integrals, they are determined by the local facet index of a given cell.

For an Expression to be considered a facet expression, the evaluation points has to be of the topological dimension of the facet

PR #836, opened 2026-06-01
Consistent orientation of output of `Expression` on facets

jorgensd:
After an intense debug session with @mscroggs, that outperforms Claude on any day, we found this funky bug in FFCx that expressions evaluated on facets were missing a permutation.

The logic behind this is:

Consider a `triangle` with local vertices

```bash
1 ---- 2
|    /
|  /
0
```
and we would like to evaluate an expression on local facet 0
`1 --- 2`.
Then, a point (x=0.2) could either
```bash
Perm 0
1 -x---- 2
```
or 
```bash
Perm 1
1 ----x- 2
```
depending of the facet_permutation (which yields a globally consistent orientation). This was disregarded in the first implementation (https://github.com/FEniCS/ffcx/pull/672)

### DIFF
--- a/ffcx/ir/elementtables.py
+++ b/ffcx/ir/elementtables.py
@@ -439,6 +439,7 @@ def build_optimized_tables(
             integral_type == "interior_facet"
             or integral_type == "ridge"
             or (is_mixed_dim and codim == 0)
+            or (integral_type == "expression" and entity_type == "facet")
         ):
             if entity_type == "facet":
                 if tdim == 1 or codim == 1:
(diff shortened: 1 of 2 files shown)


ffcx/ir/elementtables.py
```diff
@@ -439,6 +439,7 @@ def build_optimized_tables(
             integral_type == "interior_facet"
             or integral_type == "ridge"
             or (is_mixed_dim and codim == 0)
+            or (integral_type == "expression" and entity_type == "facet")
```
mscroggs, line 442: We should try to simplify the logic in this if statement, but I think we merge this fix now and revisit this later

mscroggs (approved):

ffcx/ir/elementtables.py
mscroggs: I opened #837 to remind us to look at this at some point

Issue #715, opened 2024-09-16
Casting warnings on Windows


jhale:
Originally reported by @minrk .

<When running the tests on Windows>, forgot to mention, this warning came up:

```
libffcx_forms_1b9958e91fd54e9c55f60346c28378fb63cad26e.c(722): warning C4305: 'initializing': truncation from 'double' to 'const float'
```

104,127 times when running the tests. I'm not sure if that's something that should be fixed in ffcx.

compiler warnings summary from running the tests with `pytest -vs`:

```
# grep -E -o 'warning C.+' dolfinx-win64.txt | sort | uniq -c
1665 warning C4113: 'ufcx_tabulate_tensor_float32 (__cdecl **)' differs in parameter lists from 'void (__cdecl **)(float *,const float *,const float *,const float *,const int *,const uint8_t *)'
1665 warning C4113: 'ufcx_tabulate_tensor_float64 (__cdecl **)' differs in parameter lists from 'void (__cdecl **)(double *,const double *,const double *,const double *,const int *,const uint8_t *)'
 165 warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
104127 warning C4305: 'initializing': truncation from 'double' to 'const float'
```

PR #838, opened 2026-06-10
Allow user to see compiler output in CI

jhale:
Related to issue #715.


jhale:
Unable to reproduce #715 in the CI - different/stricter injected compiler flags on condaforge?

jhale:
Going to merge this anyway, useful to have the compiler output visible.

PR #846, opened 2026-07-05
Support `numba>=0.66.0` and tidy optional dependency

schnellerhase:
- `numba.types` has been a redirect to `numba.core.types` for 6 years https://github.com/numba/numba/commit/09a570e1efe6fb45c01fc8dc96d64a799ab879ee, replaced - caught by type checkers.
- numba related functionality is now consistently not declared if `import numba` fails (`numba_ufcx_kernel_signature` was previously defined and raised, while others not defined)

### DIFF
--- a/ffcx/codegeneration/utils.py
+++ b/ffcx/codegeneration/utils.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2026 Michal Habera, Chris Richardson, Garth N. Wells and Paul T. Kühner
+# Copyright (C) 2020-2024 Michal Habera, Chris Richardson and Garth N. Wells
 #
 # This file is part of FFCx.(https://www.fenicsproject.org)
 #
--- a/ffcx/codegeneration/utils.py
+++ b/ffcx/codegeneration/utils.py
@@ -5,7 +5,6 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 """Utilities."""
 
-
 import numpy as np
 import numpy.typing as npt
 
@@ -13,6 +12,8 @@ try:
     import numba
 except ImportError:
     numba = None
+
+
 def dtype_to_c_type(dtype: npt.DTypeLike | str) -> str:
     """For a NumPy dtype, return the corresponding C type.
 
--- a/ffcx/codegeneration/utils.py
+++ b/ffcx/codegeneration/utils.py
@@ -95,10 +95,10 @@ if find_spec("numba") is not None:
         """
 
         def codegen(context, builder, signature, args):
-            null_ptr = context.get_constant(numba.types.voidptr, 0)
+            null_ptr = context.get_constant(types.voidptr, 0)
             return null_ptr
 
-        sig = numba.types.voidptr()
+        sig = types.voidptr()
         return sig, codegen
 
     @numba.extending.intrinsic
@@ -122,7 +122,7 @@ if find_spec("numba") is not None:
             the raw data pointer to the first element of the of the contiguous block of memory
             of the NumPy array to void*.
         """
-        if not isinstance(arr, numba.types.Array):
+        if not isinstance(arr, types.Array):
             raise TypeError("Expected a NumPy array")
 
         def codegen(context, builder, signature, args):
@@ -143,8 +143,8 @@ if find_spec("numba") is not None:
             """
             [arr] = args
             raw_ptr = numba.core.cgutils.alloca_once_value(builder, arr)
-            void_ptr = builder.bitcast(raw_ptr, context.get_value_type(numba.types.voidptr))
+            void_ptr = builder.bitcast(raw_ptr, context.get_value_type(types.voidptr))
             return void_ptr
 
-        sig = numba.types.voidptr(arr)
+        sig = types.voidptr(arr)
         return sig, codegen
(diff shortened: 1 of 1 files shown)


michalhabera:
> remove try-catch optional dependency management, use importlib.

Reasons for this? I don't think we use find_spec elsewhere and I prefer consistency. In fact, I am not even sure why is there a need to check if numba imports. All numba-related does not live in the top-level statements. Any anyone calling numba function must be aware that it will raise if he has no numba.

schnellerhase:
I maintained the non-definition of the functionality as was done before https://github.com/FEniCS/ffcx/pull/846/changes#diff-711cf1155ec25173aa2486ae8da736b56493d0bcd1c1ed2c5786a77f7ab4ebb5L92 with the try-catch approach by using the dedicated import lib functionality for this. I agree we should just define these functions and raise inside, but that was not like this before so I left that unchanged.

michalhabera:
> I maintained the non-definition of the functionality as was done before https://github.com/FEniCS/ffcx/pull/846/changes#diff-711cf1155ec25173aa2486ae8da736b56493d0bcd1c1ed2c5786a77f7ab4ebb5L92 with the try-catch approach by using the dedicated import lib functionality for this. I agree we should just define these functions and raise inside, but that was not like this before so I left that unchanged.

I think we agree the proper fix is to postpone the `numba` import. So why don't you just fix the `numba` API change and we can clean-up properly in a different PR?

jhale:
My understanding of the current (non-optimal) design is related to using `numba` decorators e.g. `@numba.extending.intrinsic`.

I'm not a big fan of any pattern where different installs can have more/less definitions depending on what's installed - it's best that users get an `ImportError` on runtime, rather than missing definitions. That's much easier for them to then fix (install numba).

michalhabera:
> My understanding of the current (non-optimal) design is related to using `numba` decorators e.g. `@numba.extending.intrinsic`.
> 
> I'm not a big fan of any pattern where different installs can have more/less definitions depending on what's installed - it's best that users get an `ImportError` on runtime, rather than missing definitions. That's much easier for them to then fix (install numba).

Yeah, we all agree that we should not do any magic based on availability of a dependency. We can move the intrinsic decorators and all numba-related code to a separate numba submodule file in FFCx, so users must explicitly import. But this PR should just update the old API, as @schnellerhase mentioned.

ffcx/codegeneration/utils.py
```diff
@@ -1,18 +1,15 @@
-# Copyright (C) 2020-2024 Michal Habera, Chris Richardson and Garth N. Wells
+# Copyright (C) 2020-2026 Michal Habera, Chris Richardson, Garth N. Wells and Paul T. Kühner
 #
 # This file is part of FFCx.(https://www.fenicsproject.org)
 #
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 """Utilities."""
 
+from importlib.util import find_spec
```
schnellerhase:
```suggestion
```

ffcx/codegeneration/utils.py
```diff
@@ -1,18 +1,15 @@
...
+
 import numpy as np
 import numpy.typing as npt
 
-try:
-    import numba
-except ImportError:
-    numba = None
-
```
schnellerhase, line 16:
```suggestion
try:
    import numba
except ImportError:
    numba = None
```

ffcx/codegeneration/utils.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2020-2024 Michal Habera, Chris Richardson and Garth N. Wells
+# Copyright (C) 2020-2026 Michal Habera, Chris Richardson, Garth N. Wells and Paul T. Kühner
```
michalhabera: I really dislike these Copyright statements. There is not a clear line defined for what changes constitute a copyright, and the only true record lives in git.

michalhabera (changes requested):
Wait, the official API recommends to use `from numba import types`, https://numba.readthedocs.io/en/stable/extending/interval-example.html#creating-a-new-numba-type

And the PR you've linked https://github.com/numba/numba/pull/5676 explicitly says "Restore numba.types as public API".

What am I missing here?

ffcx/codegeneration/utils.py
schnellerhase: I agree this is with the reversion strictly maintenance. reverted.

schnellerhase:
The problem is the redirect of `numba.types`. The redirect breaks `mypy`'s ability to look up the right declarations. My understanding from the PR is, `numba.types` is only maintained for legacy reasons. It was previously removed, breaking API compatibility and restored in the mentioned PR, see https://github.com/numba/numba/issues/5666.

michalhabera:
> The problem is the redirect of `numba.types`. The redirect breaks `mypy`'s ability to look up the right declarations. My understanding from the PR is, `numba.types` is only maintained for legacy reasons. It was previously removed, breaking API compatibility and restored in the mentioned PR, see [numba/numba#5666](https://github.com/numba/numba/issues/5666).

Here is a newer reference to use the public API, `import numba.types`, https://github.com/numba/numba/issues/7810

schnellerhase:
You are right; the issue is not `numba.types`, however the previous explicit forcing of the *module* import `numba.types` - which is a (hacked) redirect see https://github.com/numba/numba/blob/3f6c0dbf9f5a41516093e56ef4777bf28284d46a/numba/types/__init__.py. However `numa.types` (non-module) is pythonicly forwarded https://github.com/numba/numba/blob/main/numba/__init__.py#L81. So changing our import from  `import numba.types as types` to simply using `numba.types` fixes mypy and uses base level import. Pfew, that all for imports 😅.

michalhabera (approved):

PR #850, opened 2026-08-05
Ruff formatting fixes and checking that we can have constants from other meshes in the form

jorgensd:
Originally reported by @RemDelaporteMathurin and @ee-nn at https://github.com/festim-dev/FESTIM/pull/1216#issuecomment-5169213448

Context:
Some times (especially on manifolds), you would like to use `dx(submesh)` to get the correct spatial deriviatives of
a Coefficient or Argument defined on the submesh.
In these cases, you might want to include a constant defined on the parent domain. As Constants in DOLFINx shouldn't really have a mesh (as it serves no point, ref https://github.com/FEniCS/dolfinx/issues/4006, https://github.com/FEniCS/ufl/issues/313). The modification below allows you to use a constant from any mesh in your variational form.

### DIFF
--- a/ffcx/ir/integral.py
+++ b/ffcx/ir/integral.py
@@ -175,7 +175,7 @@ def _compute_integral_ir(
     coeffs_and_arguments = ufl.algorithms.analysis.extract_arguments(
         expression
     ) + ufl.algorithms.analysis.extract_coefficients(expression)
-    domains = [ufl.domain.extract_unique_domain(arg) for arg in coeffs_and_arguments]
+    domains = map(ufl.domain.extract_unique_domain, coeffs_and_arguments)
     for domain in domains:
         assert domain is not None
         if domain.topological_dimension != cell.topological_dimension:
--- a/ffcx/ir/integral.py
+++ b/ffcx/ir/integral.py
@@ -169,9 +169,14 @@ def _compute_integral_ir(
         if is_modified_terminal(v["expression"])
     }
 
-    # Check if we have a mixed-dimensional integral
+    # Check if we have a mixed-dimensional integral.
+    # We allow for constants to be defined on other meshes, as they require no mapping.
     is_mixed_dim = False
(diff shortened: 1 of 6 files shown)


RemDelaporteMathurin (commented):
Thanks for this @jorgensd ! So I understand that this would allow `Constant` to be defined on a parent mesh, I assume `fem.Function` on the parent mesh would still be an issue?

jorgensd:
> Thanks for this @jorgensd ! So I understand that this would allow `Constant` to be defined on a parent mesh, I assume `fem.Function` on the parent mesh would still be an issue?

Yes, this would allow for the constant to be defined on the parent mesh.
There is no easy way of integrating over `coeff_parent * dx(domain = submesh)` for codim-1 cases (codim=0 should work), as the restriction of coeff_parent to the submesh is not nicely defined in FFCx (it was how MeshView was done by @cdaversin in legacy FEniCS, but that had some other drawbacks).

ffcx/ir/integral.py
```diff
@@ -169,9 +169,15 @@ def _compute_integral_ir(
...
 
-    # Check if we have a mixed-dimensional integral
+    # Check if we have a mixed-dimensional integral.
+    # We allow for constants to be defined on other meshes, as they require no mapping.
     is_mixed_dim = False
-    for domain in ufl.domain.extract_domains(expression):
+    coeffs_and_arguments = ufl.algorithms.analysis.extract_arguments(
+        expression
+    ) + ufl.algorithms.analysis.extract_coefficients(expression)
+    domains = [ufl.domain.extract_unique_domain(arg) for arg in coeffs_and_arguments]
```
schnellerhase:
```suggestion
    domains = map(ufl.domain.extract_unique_domain, coeffs_and_arguments)
```

schnellerhase (commented):
Can you add a test for the use-case that this enables?

PR #852, opened 2026-08-16
Split quadrature loop into evaluation and contraction

garth-wells:
Splits the quadrature loop into (i) evaluation of the varying quantities and (ii) the tensor contraction, so that math functions of `SpatialCoordinate` can vectorise. Worth 2.4x on a P2 Poisson load-vector kernel and 1.87x on end-to-end RHS assembly in DOLFINx.

## Problem

The load vector `L = inner(f, v)*dx(degree=p+1)` with the usual manufactured source

```python
f = 3.0 * ufl.pi**2 * ufl.sin(ufl.pi*x[0]) * ufl.sin(ufl.pi*x[1]) * ufl.sin(ufl.pi*x[2])
```

compiles to scalar `sin` calls at P2 but vectorised ones at P1 (`-O3 -ffast-math -march=native`, GCC 13.3):

| element | generated object references |
| --- | --- |
| P1 | `_ZGVdN4v_sin` — libmvec, four doubles per call |
| P2 | `sin` — scalar `__sin_fma`, 15 calls per cell |

`perf` on the P2 RHS assembly loop (tetrahedra, N=64, 1.57M cells, 0.183 s per assembly): `__sin_fma` 61.5%, `sin@plt` 2.6%, rest of `tabulate_tensor` 23.3%, DOLFINx assembler 11.5%.

GCC on the P2 kernel:

```
missed: couldn't vectorize loop
missed: not vectorized: no vectype for stmt: _338 = FE2_C0_Q175[0][0][iq_192][0];
```

The generated loop interleaves the two stages:

```c
for (int iq = 0; iq < 5; ++iq)
{
  // three coordinate interpolations from FE2_C0_Q175[1][1][5][4]   (stride 4)
  double sv_175_1 = sin(sv_175_0);          // wants to be vectorised
  ...
  fw0 = sv_175_9 * weights_175[iq];
  for (int i = 0; i < 10; ++i)
    A[i] += fw0 * FE0_C0_Q175[0][0][iq][i];  // stride 10
}
```

The strided basis-table reads in the contraction are what GCC declines to vectorise, and they sit in the same loop body as the `sin` calls, so the whole loop stays scalar. At P1 the basis and coordinate elements coincide, giving a single `[1][1][4][4]` table with stride 4 — exactly the vector width — and the loop vectorises.

Ruled out on the real kernel: the trip count (forcing `iq < 4`, `< 6`, `< 8` still scalar), the cost model (`-fvect-cost-model=unlimited` unchanged), the quadrature rule (basix gives 5 points at degree 3 where FIAT gives 6, so FFCx already evaluates fewer sines), and `#pragma omp simd` with `-fopenmp-simd` (object still references scalar `sin`). Three one-line perturbations each flip it to vectorised, showing how marginal the decision is: pointing the contraction at the stride-4 table, `sin` -> `sqrt`, or dropping the `weights[iq]` factor.

## Change

`fw` intermediates are stored in a `double[nq]` temporary in the first loop and read back in the second. Applies only to single-index (non tensor-product) rules with `fw` intermediates; everything else keeps the fused loop.

## Results

Kernel timed directly, 1.57M cells, cache-resident coordinates:

| kernel | time | per cell | |
| --- | --- | --- | --- |
| fused loop (main) | 0.1551 s | 98.6 ns | |
| split loop (this PR) | 0.0663 s | 42 ns | **2.4x** |

End to end in DOLFINx, P2 N=64, serial, best of 3:

| | main | this PR | |
| --- | --- | --- | --- |
| RHS assemble | 0.1829 s | 0.0978 s | **1.87x** |
| RHS assemble, BC applied | 0.2295 s | 0.1489 s | 1.54x |
| matrix re-assemble | 1.9136 s | 1.8776 s | 1.02x |
| matrix create + assemble | 4.3560 s | 4.2826 s | 1.02x |

No regression at P1 (N=100, 6M cells): RHS 1.00x, RHS+BC 0.99x, matrix re-assemble 0.97x, create+assemble 0.98x — all within run-to-run noise.

## Correctness

`||b||` and `||A||_F` identical to all printed digits at P1, P2 and P3 against a build of this branch's parent; in the P2 benchmark `||A||_F = 85.6547612828181` and 60 859 905 nonzeros, unchanged. `python -m pytest test/` passes: 212 passed, 1 skipped (`test_numba.py` and `test_signatures.py` not run — no numba in my environment).

## Open questions

- Split unconditionally, or only when the integrand contains a `MathFunction`? The split costs a `double[nq]` temporary and an extra loop. The P1 and matrix numbers above suggest it is neutral otherwise, but that is two data points.
- Interaction with the existing `licm` and section-fusing passes deserves a look, as does extending it to tensor-product rules, which this skips.
- Only GCC 13.3 was tested; whether the same shape helps clang is unknown.

Environment: GCC 13.3.0 (Ubuntu 24.04), Xeon w7-3445 (AVX-512), DOLFINx `91e71e2`.

AI assistance: I used Claude Code (Opus 5) to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.


jorgensd:
Could we test this for an "unreasonable" high quadrature degree, for instance 30 in 2D and 15 in 3D, which is the highest degree we have special rules for: https://github.com/FEniCS/basix/blob/ce3db44e2a02435a972446ebff07d32889a43346/cpp/basix/quadrature.cpp#L4936-L4951

As we are looking at P1 and P2 kernels here, the estimated degree is likely 4 or 5 for the math functions at max.

garth-wells:
This PR is mainly just to keep track of an issue that came up during profiling. I haven't looked closely.

garth-wells:
Independent verification of the numbers in this PR, using an interleaved A/B:
base and patch alternated within each round, FFCx reinstalled and the FFCx cache
cleared for each measurement, best of three repetitions per measurement. DOLFINx
assembly on a unit cube of tetrahedra, serial, Xeon w7-3445. All norms
bit-identical between base and patch.

Interleaving matters here: on this machine non-interleaved A/B drifts by up to 8%
between runs, enough to invent or hide effects below ~1.2×.

**The headline claim holds.** P2 load vector, `sin(pi x) sin(pi y) sin(pi z)`
source, 1.57M cells, `-O3 -ffast-math -march=native`:

| round | main | this PR | ratio |
| --- | --- | --- | --- |
| 1 | 0.1750 s | 0.0944 s | 1.85 |
| 2 | 0.1784 s | 0.0945 s | 1.89 |
| 3 | 0.1787 s | 0.1032 s | 1.73 |

0.175 → 0.094 s matches the 0.183 → 0.098 s quoted in the commit message.

**The secondary claim holds too.** P1 matrix re-assembly, 6M cells, same flags:
0.985 / 1.027 / 1.021 — unchanged, as stated.

**One qualification worth adding to the description.** The win requires
`-ffast-math` (or at least `-fno-math-errno`). At FFCx's default compile flags,
where generated code is built with the interpreter's CFLAGS (`-O2`), the same P2
RHS case measures 1.01 / 0.97 / 1.05 — no effect. That is expected rather than a
problem: GCC only vectorises calls to `sin` onto libmvec when it is allowed to,
so at `-O2` there is no vectorisation for the split to unblock. Users on the
default configuration see neither gain nor loss.

**Composes with #853.** That PR restructures the "Tensor Computation" section
this one separates out, so they interact. Cherry-picking #853's three commits on
top of this branch merges cleanly (one auto-merge in `integral_generator.py`),
`test/` passes 211, and both wins survive:

| case | base | combined | ratio |
| --- | --- | --- | --- |
| P2 RHS, fast-math | main 0.176 s | 0.096 s | 1.83 preserved |
| P1 matrix, default flags | main 3.04 / 2.80 s | 1.66 / 1.60 s | 1.84 / 1.75 preserved |

Against this branch alone the combined tree costs ~2–5% on the RHS under
fast-math (0.954 / 0.984), which is inside the noise band above.

The two changes are complementary: this one wins under fast maths and is neutral
at the default flags; #853 wins at the default flags and is neutral under fast
maths.

garth-wells:

PR #853, opened 2026-08-17
Make the tensor contraction loop cheaper: group terms, accumulate once, nest contiguously

garth-wells:
Three related changes to the generated tensor contraction. Each is a separate
commit and each was measured on its own.

## 1. Group terms sharing an inner-loop factor

Loop-invariant code motion hoisted each term of an entry separately, creating one
temporary array and one accumulation per term. A 3D Poisson operator has one term
per pair of derivative directions, so the inner loop carried nine temporaries
where three suffice: terms sharing the factor left in the inner loop differ only
in what was hoisted, so one temporary can hold their sum.

## 2. Accumulate into each entry once

After grouping, each entry was still updated once per group — three
read-modify-writes of the same address per inner iteration. Sum the group
contributions and assign once.

## 3. Nest with the contiguous tensor index innermost

The tensor indices were reversed before building the loop nest, putting the row
index of the element tensor in the inner loop, so each iteration stepped over a
whole row: a stride of 4 for P1 tetrahedra and 20 for P3. That stops the
accumulation vectorising and leaves the hoisted temporaries indexed by the inner
loop. Nesting in the natural order walks the element tensor contiguously and
makes the temporaries loop-invariant.

Together, for P1 Poisson on tetrahedra, the contraction goes from nine
temporaries and nine accumulations:

```c
double temp_0[4] = {0};
for (int j = 0; j < 4; ++j) temp_0[j] = fw0 * FE0_C0_D100_Q421[0][0][0][j];
/* ... temp_1 .. temp_8, nine arrays, nine loops ... */
for (int j = 0; j < 4; ++j)
  for (int i = 0; i < 4; ++i) {
    A[4 * (i) + (j)] += FE0_C0_D100_Q421[0][0][0][i] * temp_0[j];
    A[4 * (i) + (j)] += FE0_C0_D100_Q421[0][0][0][i] * temp_1[j];
    /* ... nine accumulations, A strided by 4 ... */
  }
```

to three, with a contiguous single accumulation:

```c
double temp_0[4] = {0};
for (int i = 0; i < 4; ++i)
  temp_0[i] = fw0 * FE0_C0_D100_Q421[0][0][0][i]
            + fw1 * FE1_C1_D010_Q421[0][0][0][i]
            + fw2 * FE1_C2_D001_Q421[0][0][0][i];
/* temp_1, temp_2 */
for (int i = 0; i < 4; ++i)
  for (int j = 0; j < 4; ++j)
    A[4 * (i) + (j)] += FE0_C0_D100_Q421[0][0][0][j] * temp_0[i]
                      + FE1_C1_D010_Q421[0][0][0][j] * temp_1[i]
                      + FE1_C2_D001_Q421[0][0][0][j] * temp_2[i];
```

## Performance

DOLFINx matrix re-assembly, unit cube of tetrahedra, serial, Xeon w7-3445.
Interleaved A/B (base and patch alternated within each round, FFCx reinstalled
each time), best of three repetitions per measurement. Frobenius norm of the
assembled operator bit-identical in every run.

At FFCx's default compile flags, where generated code is built with the
interpreter's CFLAGS (`-O2`):

| case | commit 1 | commits 2+3 | cumulative |
| --- | --- | --- | --- |
| P1, 6M cells | 2.84 s → 1.79 s (1.52 / 1.64 / 1.62) | 1.78 s → 1.61 s (1.085 / 1.086 / 1.087) | **2.84 s → 1.61 s, 1.8x** |
| P3, 1.57M cells | 179 s → 65 s (2.67 / 2.84) | 66.3 s → 44.1 s (1.507 / 1.500) | **179 s → 44.1 s, 4.1x** |

Under `-O3 -ffast-math -march=native` all three are neutral — the compiler is
then free to do the regrouping itself, and the small element matrices are already
in L1 so the stride costs little:

| case | commit 1 | commits 2+3 |
| --- | --- | --- |
| P1, 6M cells | 1.03 / 1.00 / 1.01 | 1.008 / 1.034 / 0.990 |
| P3, 1.57M cells | 1.04 / 1.01 | — |

So this helps users on the default configuration and costs nothing to users who
opt into fast maths. Non-interleaved measurement on this machine drifts by up to
8% between runs, so the neutral rows should be read as "no change" rather than as
a small gain.

For context on where the remaining kernel time is: at default flags the kernel is
roughly three-quarters of P3 assembly (44 s total, of which PETSc insertion is
~10 s and flag-independent), and switching that case to
`-O3 -ffast-math -march=native` takes it to 12 s. Changing the default compile
flags is worth more than these changes combined.

## Correctness

`test/` passes: 211 passed, 1 skipped. (`test_cmdline_simple` fails in my
environment because the `ffcx` CLI is not on the test subprocess's PATH; it fails
identically on the branch point.)

Change 3 alters which axis the grouping runs over, so the `fw` factors attach to
different groups. I checked all nine terms of the P1 operator regroup correctly:
`A[i][j]` is unchanged, it is not a transpose. Worth stating because Poisson is
symmetric and would hide a transpose in the Frobenius norm — the non-symmetric
forms in the test suite would not, and they pass.

Note that changes 1 and 2 regroup distributively, `a*b + a*c -> a*(b+c)`, so
results may in principle differ in the last ulp. Every case measured here was
bit-identical, but that is not guaranteed in general.

## Notes

Draft: I would like a second opinion on whether the last-ulp caveat is
acceptable, whether the grouping key in `_key` should be replaced by proper
structural equality on LNodes expressions, and whether removing the index
reversal in `integral_generator.py` has a rationale I have missed — it carried
only the comment `# reverse B_indices` and no test depends on it.


garth-wells:

PR #859, opened 2026-08-19
Only split the quadrature loop when it has a math function to vectorise

garth-wells:
Stacked on #852 — targets that branch, not `main`, since this is a small refinement of its own logic. Retarget to `main` once #852 merges.

## Problem

#852 splits the quadrature loop into an evaluation pass and a contraction pass whenever any `fw` intermediate exists, so vectorised math-library calls (`sin`, `exp`, ...) can escape the strided contraction loop. The condition is broader than the mechanism: an `fw` intermediate exists for nearly every integral, math function or not. A bare mass matrix has one too — just the quadrature weight times `|det J|`:

```c
double fw0_q[14];
for (int iq = 0; iq < 14; ++iq)
{
  double fw0 = 0;
  fw0 = sp_d8a_18 * weights_d8a[iq];   // nothing here can vectorise
  fw0_q[iq] = fw0;                     // stored...
}
for (int iq = 0; iq < 14; ++iq)
{
  double fw0 = fw0_q[iq];              // ...then reloaded, for no reason
  ...
```

Two loops, an extra array, and a store/load round-trip through memory, for zero vectorisation gain.

## Benchmarks

Building on this branch, I generated and directly timed kernels (self-calibrated microbenchmark, no DOLFINx dependency) for ten forms spanning scalar/vector/mixed Lagrange, N1curl, SIPG DG, and the HyperElasticity demo's nonlinear residual + Jacobian. The unconditional split regresses every math-function-free integrand in that set:

| kernel | main | #852 (unconditional split) | + this gate |
| --- | --- | --- | --- |
| mass matrix, P2 tet, `-O2` | 779.2 ns | 877.4 ns (**0.89x**) | 789.1 ns (**0.99x**) |
| mass matrix, P2 tet, `-O3 -ffast-math -march=native` | 523.2 ns | 567.2 ns (**0.92x**) | 584.0 ns (0.90x, noise-level) |
| load-vector with `sin`, P2 tet, `-O2` (#852's own target) | 533.4 ns | 519.3 ns (**1.03x**) | 528.1 ns (1.01x — split still applies, gate correctly lets it through) |

No regression anywhere else in the ten-form set at `-O2`; #852's own wins are untouched (the gate is a no-op whenever a math function is genuinely present).

## Fix

Gate the split on the evaluation code actually containing a call to a `MathFunction`, matching the open question #852's own description raised (*"split unconditionally, or only when the integrand contains a MathFunction?"*). A small recursive walk over the LNodes statement tree, checked once before committing to the split — see the diff.

## Correctness

`pytest test/` (excluding `test_jit_forms.py`/`test_numba.py`/`test_signatures.py`, which need a numba/numpy combination not in my environment): 46 passed, 1 skipped, unchanged from this branch without the fix.

## Note

A wider nonlinear form (HyperElasticity's residual) still regresses under the split even with this gate applied (it does contain a math function, `exp`), likely because the split's overhead scales with the number of live `fw` intermediates and the vectoriser's cost model doesn't turn a profit on a loop body that large. Left open rather than patched with a second heuristic — flagging here in case it's useful context for review, happy to open a separate issue.

AI assistance: I used Claude Code (Opus 5) to draft this change, building on the prior #852 work. I reviewed and take responsibility for the final contribution.

### DIFF
--- a/ffcx/codegeneration/integral_generator.py
+++ b/ffcx/codegeneration/integral_generator.py
@@ -48,6 +48,31 @@ def extract_dtype(v, vops: list[Any]):
     return L.merge_dtypes(dtypes)
 
 
+def _contains_mathfunction(node) -> bool:
+    """Search an LNodes statement/expression tree for a call to a MathFunction."""
+    if isinstance(node, list):
+        return any(_contains_mathfunction(n) for n in node)
+    if isinstance(node, L.MathFunction):
+        return True
+    if isinstance(node, (L.StatementList, L.Section)):
+        return _contains_mathfunction(node.statements)
+    if isinstance(node, L.ForRange):
+        return _contains_mathfunction(node.body)
+    if isinstance(node, L.VariableDecl):
+        return node.value is not None and _contains_mathfunction(node.value)
+    if isinstance(node, (L.ArrayDecl, L.Comment)):
+        return False
+    if isinstance(node, L.Statement):
+        return _contains_mathfunction(node.expr)
+    if isinstance(node, L.NaryOp):
+        return any(_contains_mathfunction(a) for a in node.args)
+    if isinstance(node, L.BinOp):
+        return _contains_mathfunction(node.lhs) or _contains_mathfunction(node.rhs)
+    if isinstance(node, L.PrefixUnaryOp):
+        return _contains_mathfunction(node.arg)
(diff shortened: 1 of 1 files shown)


garth-wells:

PR #861, opened 2026-08-19
Wire up reciprocal-CSE, and fold power-of-two round-trips

garth-wells:
## Summary

Two small, verified-correct simplifications to the scalar geometry setup FFCx emits for every affine-cell kernel:

1. **Reciprocal-CSE, actually wired in.** An earlier benchmarking pass (see the report linked below) measured this by hand-patching generated code, but never landed it — tracing where it belongs turned up why: the repeated per-entry divisions (every cofactor of an affine cell's pseudo-inverse Jacobian dividing by the same determinant) live in `generate_partition`'s `intermediates` list, which `optimize()` never touched (only the separate `definitions` list was optimised). Fixed by running a reciprocal-hoisting transform directly on `intermediates`, for both the piecewise (once-per-cell) and varying (per-quadrature-point) partitions.

2. **A new lead: power-of-two round-trips.** Symmetric-gradient forms (e.g. plain elasticity) showed a further, related waste: a value doubled in one place and halved in another, landing back on the original value, computed a second time under a different name — because ordinary CSE only recognises syntactically identical expressions, not algebraic identities. Multiplying and dividing by an exact power of two never rounds in IEEE-754 (`(x + x) * 0.5` is bit-identical to `x` for any finite, non-subnormal `x`), so this is always safe to fold, but GCC won't do it itself without `-ffast-math`. Added `power_of_two_cse_statements`, tracking each declared scalar's value as `(base symbol, power-of-two exponent)` through chains of `+self` and multiply/divide-by-literal-power-of-two, aliasing a statement whose value is already known under an earlier name instead of recomputing it. Must run *before* reciprocal-CSE, which would otherwise hide the literal power-of-two divisors this looks for behind a reciprocal symbol.

## Where this came from

Full writeup, including the benchmark harness bug this survey also found and fixed, and an honest account of a benchmark-host instability that this session ran into partway through measuring these two changes: https://claude.ai/code/artifact/b2b2e22b-e0e5-4803-83ea-960ae2bc1506 (§9, "Two verified codegen fixes, and a wall this session hit").

The short version on performance: both changes are real, verified-correct, essentially risk-free. Whether they meaningfully close FFCx's gap on vector-valued and mixed forms (elasticity, vector Poisson, Taylor-Hood Stokes) is **not conclusively established** — the cleanest same-session measurement (reciprocal-CSE alone, before the host destabilised) landed close to neutral under `-O3 -march=native` specifically, consistent with GCC's own SLP vectoriser already capturing much of what hand-hoisting a reciprocal buys once the vectoriser is switched on at all (the original hand-patched measurement was `-O2`-only, with no vectoriser in the picture). Recommend landing on correctness/risk grounds regardless, and re-benchmarking on a stable host before drawing further performance conclusions.

## Correctness

- Full test suite: 46 passed, 1 skipped (unchanged from main).
- Numerically cross-checked old vs. new generated code for `elasticity_p1`'s bilinear kernel directly: compiled both as shared libraries, called via `ctypes` with identical non-degenerate tetrahedron coordinates, diffed the output element matrices — max absolute difference `2.22e-16` (one part in 2^52, a single extra rounding step from `x * (1/d)` vs `x / d`, not a correctness issue — the expected order of difference any unconditional version of the reciprocal trick produces).


garth-wells:

PR #862, opened 2026-08-19
Unroll coordinate-dofs linear combination for non-tensor-product geometry

garth-wells:
## Summary

`_define_coordinate_dofs_lincomb()` (used for both the Jacobian and `SpatialCoordinate`) generated the coordinate-dofs sum as a genuine runtime `for` loop over the coordinate element's dofs. For a P1 tetrahedron that's a 4-iteration reduction — small enough that GCC's auto-vectoriser mis-vectorises it into a permute-then-horizontal-reduce sequence (`vpermi2pd`/`vunpckhpd`/`vextractf64x2`/`valignq`) that costs more than the four scalar multiply-adds it replaces. A hand-written kernel would never emit this as a loop at all — it would unroll the identical sum into direct symbolic subtractions at the source level, which is also easier for GCC to constant-fold against a sparse basis-derivative table (P1's table is mostly `0.0`/`±1.0`).

This unrolls the sum into a single literal-indexed expression at codegen time instead. Guarded by `tabledata.has_tensor_factorisation`: tensor-product/high-order hex/quad coordinate elements (which can have many more dofs, where a runtime loop is the right call) keep the original loop-based code path unchanged.

## Measured effect

Compiled instruction count for `poisson_p1`'s bilinear kernel dropped 260→209 (GCC 13, `-O3 -march=native -mprefer-vector-width=512 -ffp-contract=fast`); every vectoriser shuffle/reduce instruction is gone from the disassembly. Benchmarked the same kernel, same flags, before/after:

| form | before | after |
|---|---|---|
| poisson_p1 | 54.95 ns | 45.90 ns |

(elasticity_p1/mass_p2 also touch this code path; their overall movement in the same session was dominated by an unrelated GCC 13→14 compiler swap, tracked separately.)

## Test plan

- [x] `pytest` — 212 passed, 1 skipped (including the tensor-product suite, which exercises the untouched fallback path for tensor-factorised coordinate elements)
- [x] Numeric correctness spot-checked via ctypes (P1 stiffness matrix row sums ≈ 0)
- [x] Disassembly-verified: no vectoriser shuffle/reduce instructions remain for the affected kernels

### DIFF
--- a/ffcx/codegeneration/definitions.py
+++ b/ffcx/codegeneration/definitions.py
@@ -210,14 +210,43 @@ class FFCXBackendDefinitions:
         if mt.restriction == "-":
             offset = num_scalar_dofs * dim
 
-        code = []
-        declaration = [L.VariableDecl(access, 0.0)]
-        body = [L.AssignAdd(access, dof_access[ic.global_index * dim + begin + offset] * FE)]
-        code = [L.create_nested_for_loops([ic], body)]
+        if tabledata.has_tensor_factorisation:
+            # Tensor-product dof counts can be large (high-order hex/quad
+            # geometry); keep the general runtime-loop path.
+            code = []
+            declaration = [L.VariableDecl(access, 0.0)]
+            body = [L.AssignAdd(access, dof_access[ic.global_index * dim + begin + offset] * FE)]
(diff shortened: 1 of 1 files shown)


garth-wells:

PR #863, opened 2026-08-19
Fuse and dedup licm's hoisted temps, and fuse the piecewise scale factor with the quadrature weight

garth-wells:
## Summary

Three related fixes, found while chasing a residual performance gap on several forms, all touching how the codegen backend hoists and reuses scalar factors:

**1. Fuse the per-temp hoist loops** (`licm()` in `optimizer.py`). `licm()` hoists loop-invariant factors of each block entry into a temporary array, computed once per outer-loop iteration ahead of the main accumulation loop. For each hoisted temp it emitted a separate `ForRange` -- all sharing the identical `(index, begin, end)` -- and never re-fused them afterwards. For a scalar Poisson-type form this is one or two temps, harmless. For a form with more block entries (`elasticity_p1`'s 3x3 block structure: 27 temps; `HyperElasticity`'s much larger tensors) this produced dozens of independent same-range loops back to back, each individually far too small for GCC to do anything useful with. One loop computing all of them together per iteration is both far less loop overhead and a much better vectorisation target. Also drops the dead zero-initialiser on the hoisted temp arrays.

**2. Dedup identical hoisted expressions** (`licm()`). A vector-valued form with uncoupled, identical per-component blocks (e.g. vector Poisson's diagonal Laplacian blocks -- component 0, 1, 2 all solve the identical scalar problem) produced the exact same hoisted expression once per component: recomputed and stored from scratch each time instead of once and reused. Cached by the hoisted expression's structural key (new `_key()` helper -- a hashable identity for LNode expressions, which aren't otherwise hashable) across the whole `licm()` call.

**3. Fuse the piecewise scale factor with the quadrature weight** (`integral_generator.py`). For an affine cell, the per-block scale factor (a metric-tensor entry scaled by `|det J|`) is computed once per cell, then each entry gets multiplied by the quadrature weight separately inside the quadrature loop -- 2 multiplies per entry. Fusing `|det J| * weight` into one scalar first, then multiplying each entry by that once, costs `N + Q` multiplies for N entries and Q quadrature points against the previous `2N`. The previous strategy is the right one whenever `Q > N` (the common case above the lowest quadrature orders, since the shared scale amortises across every point); it's specifically wrong for the low-point-count rules low-order affine geometry uses. Detects when every block entry has a piecewise value that's a plain two-factor product sharing one factor in common, and only fires when there are more sharing entries than quadrature points -- falls back to the existing per-entry multiply otherwise, no risk to any case it doesn't apply to. The original, now sometimes-unused piecewise declarations are cleaned up by a small generic text-level pass in `C/integral.py` that drops any side-effect-free scalar declaration never read in the rendered function body (this is also what was tripping `-Wunused-variable` in the JIT test suite once the fusion made some of them dead -- fixed generically rather than by tracking liveness through the LNode rewrite itself).

## Measured effect

GCC 14.2, `-O3 -march=native -mprefer-vector-width=512 -ffp-contract=fast`:

| form | before (this PR) | after all 3 commits |
|---|---|---|
| poisson_p1 | 45.9 ns | 42.6 ns |
| elasticity_p1 | 199.1 ns | 151.5 ns |
| vector_poisson_p2 | 564.2 ns | 517.9 ns |
| stokes_th | 796.7 ns | 733.5 ns |
| HyperElasticity (6 kernels, total) | 2146.2 us | ~2000 us |
| mass_p2, dg_sipg, load_sin (no fusion opportunity) | -- | unaffected / noise-level |

Commit 3's fusion gives `poisson_p1` most of its remaining win, since the scale-factor computation is most of that tiny kernel's total work. `elasticity_p1` and `stokes_th` see the same fusion fire but move less from it, since a much larger tensor-accumulation phase dominates their total runtime -- commits 1-2 did most of the work for those two.

## Test plan

- [x] `pytest` -- 212 passed, 1 skipped, on all three commits
- [x] Numeric spot-check of the P1 stiffness matrix via ctypes (row sums ~= 0) after commit 3
- [x] Verified via generated C source and disassembly that each rewrite compiles to noticeably fewer instructions with no correctness change

### DIFF
--- a/ffcx/codegeneration/C/integral.py
+++ b/ffcx/codegeneration/C/integral.py
@@ -6,6 +6,7 @@
 """Generate UFCx code for an integral."""
 
 import logging
+import re
 import sys
 
 import basix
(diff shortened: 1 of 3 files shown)


garth-wells:

PR #877, opened 2026-09-01
Maintain reciprocal dtype

schnellerhase:
Optimizer introduced in https://github.com/FEniCS/ffcx/pull/876 disregarded dtype of expression. Caught in type templated unit tests of the C++ backend at https://github.com/fenics-dolfiny/ffcx-backends/.


michalhabera:
@schnellerhase Just to clarify, this stored division result in the geometry computations into the scalar dtype, while it should store them into geometry dtype?

schnellerhase:
Yes that's the error we observed. However there is no constraint on the same optimisation being run on other expressions, so the dtype should be deduced from it directly.

michalhabera (approved):

garth-wells:
@schnellerhase can you add a test that would catch this?

PR #885, opened 2026-09-21
Add stable names for generated integral kernels

garth-wells:
This adds stable, user-facing names for generated C integral kernels while
retaining the existing hash-named functions as forwarding symbols.

- Derive kernel names from Python form variable names during file-based
  compilation.
- Allow an exact, explicit per-integral function name with `ffcx_kernel_name` metadata.
- Declare stable kernels in the generated header so applications can call them
  directly.
- Keep API/JIT compilation hash-based when neither an object name nor metadata
  override is provided.
- Validate identifiers, grouped-integral consistency, and uniqueness.
- Add a demo, user documentation, and focused code-generation tests.

The stable function contains the generated kernel body and is used by the
`ufcx_integral` object. This allows a caller that includes the generated source
in its translation unit to call a predictable symbol that can be inlined. The
old hash-named function forwards to the stable function for compatibility.

## Simplifications enabled

Applications that call generated kernels directly currently have to embed an
implementation detail such as:

```c
tabulate_tensor_integral_0f31b12dbe4babd413e1ac21cdf56eb73b71c30d_triangle(...);
```

With this change, a UFL file can request a stable name:

```python
a = inner(u, v) * dx(metadata={"ffcx_kernel_name": "tabulate_tensor_mass"})
```

and consuming C or C++ code can call the declaration emitted in the generated
header:

```c
tabulate_tensor_mass(...);
```

For forms without an explicit name, the Python form variable supplies the
name automatically. This means direct-kernel examples, including DOLFINx's
custom-kernel demo, no longer need to:

- copy SHA-1-derived symbols from generated source;
- update application code whenever the UFL signature, FFCx version or UFCx
  signature changes;
- inspect or parse generated files during configuration to discover a symbol;
- traverse a `ufcx_form` and call the kernel indirectly through a function
  pointer when a direct, inlineable call is required; or
- duplicate a generated kernel declaration in application code.

Including the generated source in a C++ translation unit therefore becomes a
straightforward way to expose both the generated body and a predictable direct
call to the optimiser. Existing hash-based calls remain valid through the
forwarding symbol.

## Tests

- `ruff format --check .`
- `ruff check .`
- `pytest -q test` (393 passed after rerunning the CLI test with the virtual
  environment executable on `PATH`; 1 expected optional skip)
- `pytest -q demo/test_demos.py` (224 passed, 16 expected skips)
- `python -m sphinx -W -b html doc/source ...`
- Generated-header C++20 compile check

`mypy -p ffcx` reports seven pre-existing errors and no errors in the changed
files.


jorgensd:
> ```c
> tabulate_tensor_mass_mass_triangle(...);
> ```

Why duplicate "mass" in the signature?
Also is really "mass_triangle" explicit enough?
What if you sweep over meshes with different polynomial degree, then (triangle, 1) and (triangle, 2) would get the same signature even if the coordinate elements are different.

jorgensd:
I think it should either be:
> `tabulate_tensor_mass`
or 
> `tabluate_tensor_mass_triangle_cmapdegree_cmap_lvariant`

garth-wells:
> I think it should either be:
> 
> > `tabulate_tensor_mass`
> > or
> > `tabluate_tensor_mass_triangle_cmapdegree_cmap_lvariant`

The point is that it can be whatever the user wants it to be. It's for special use case, like when including the .c file to get kernel inlining. By default it continue to use a hash that depends on the kernel, options, etc.

garth-wells:
> > ```c
> > tabulate_tensor_mass_mass_triangle(...);
> > ```
> 
> Why duplicate "mass" in the signature? Also is really "mass_triangle" explicit enough? 

Fix, such that user control the full name

The name is for the user to decide if they choose for explicit control the function name.

>What if you sweep over meshes with different polynomial degree, then (triangle, 1) and (triangle, 2) would get the same signature even if the coordinate elements are different.

PR #888, opened 2026-09-21
options: only pretty-print option values when INFO is enabled

garth-wells:
## Problem

`get_options()` builds an INFO log record unconditionally:

```python
logger.setLevel(int(options["verbosity"]))
logger.info("Final option values")
logger.info(pprint.pformat(options))
```

The default verbosity is `WARNING`, so the record is discarded — but `pprint.pformat` still runs. Measured, it is essentially the entire cost of the function:

```
pprint.pformat(options): 0.042 ms
get_options():           0.042 ms
pformat share:           100%
```

`get_options()` is called twice per `compile_forms()`, so this is ~0.084 ms on every JIT cache hit. Small in absolute terms, but a large share of what a cache hit should cost.

## Fix

Guard the two log calls with `logger.isEnabledFor(logging.INFO)`. The `setLevel` call immediately above means the guard sees the verbosity the caller asked for, so behaviour at `verbosity <= INFO` is unchanged.

## Measurements

`fem.form()` against a real 24,196-file `~/.cache/fenics`, median of 50 warm calls. Stacked on #887, which removes the cache-directory scan that otherwise dominates and hides this:

| | `fem.form` |
|---|---|
| main (`ce8c9cb`) | 9.023 ms |
| + #887 | 0.257 ms |
| + this change | **0.139 ms** |

1.85x on top of #887; 65x for the two together. The change is independent of #887 and correct on its own — it is just not visible until the directory scan is gone.

## Testing

`test_options_not_formatted_when_info_disabled` asserts `pprint.pformat` is not called at `verbosity=WARNING`; `test_options_formatted_when_info_enabled` asserts the option values still reach the log at `verbosity=INFO`. Verified as a genuine regression test: the first fails on `main` ("Expected 'pformat' to not have been called. Called 1 times.") and passes with the fix.

Full suite: 386 passed, 1 skipped (`pygraphviz` not installed, unrelated). `ruff check` and `ruff format --check` clean.


garth-wells:
Two corrections to the numbers above, from re-measuring with the configurations interleaved over three rounds rather than back to back in one pass.

**The absolute timings were inflated.** The interleaved runs give:

| | `fem.form` |
|---|---|
| #887 | 0.115 ms |
| #887 + this PR | 0.066 ms |

so 1.74x rather than the 1.85x quoted, and both baselines about half what the PR body says (0.257 / 0.139 ms). The original pair was measured immediately after a `pip install` with cold caches; at this scale run-to-run variation is large enough to matter. The direction and rough size of the effect stand.

**"`get_options()` is called twice per `compile_forms()`" is wrong.** It is called once per `compile_forms()`. What is true, and is what the measurement reflects, is that a single `fem.form()` triggers two `ffcx.options.get_options` calls and hence two `pprint.pformat` calls — confirmed by `cProfile` call counts (200 of each over 100 `fem.form` calls).

Neither changes the case for the PR.

